PR #22, opened 2020-02-14
Uri Value Component

### LINKED ISSUE #17: Viewer Value Component: UriValueComponent
Implement `BaseValueComponent` for a `UriValue`.

mdelez:
closes #17

### DIFF
--- a/projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
@@ -237,12 +237,12 @@ describe('UriValueComponent', () => {
 
     it('should set a new display value', () => {
 
-      const newInt = new ReadUriValue();
+      const newUri = new ReadUriValue();
 
-      newInt.uri = 'http://www.reddit.com';
-      newInt.id = 'updatedId';
+      newUri.uri = 'http://www.reddit.com';
+      newUri.id = 'updatedId';
 
-      testHostComponent.displayInputVal = newInt;
+      testHostComponent.displayInputVal = newUri;
 
       testHostFixture.detectChanges();
 
@@ -261,7 +261,7 @@ describe('UriValueComponent', () => {
     });
   });
 
-  describe('create an integer value', () => {
+  describe('create a URI value', () => {
 
     let testHostComponent: TestHostCreateValueComponent;
     let testHostFixture: ComponentFixture<TestHostCreateValueComponent>;
--- a/projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
@@ -1,6 +1,43 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { UriValueComponent } from './uri-value.component';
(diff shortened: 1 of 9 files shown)


tobiasschweizer:
Thanks, I try to have a look at that today.

mdelez:
I think the validator could use some work but I'm not entirely sure what Knora accepts as a valid url

tobiasschweizer:
> I think the validator could use some work but I'm not entirely sure what Knora accepts as a valid url

any valid `xsd:URI` (https://docs.knora.org/paradox/02-knora-ontologies/knora-base.html#subclasses-of-value, https://github.com/dasch-swiss/knora-api/blob/ccb64bbcf16b80d55423ff44b28c89d34bf39581/webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala#L713-L714). 

I think it would be good idea to put the validators in a central place so they could be reused. Also then the regex would only be created once and not with every instance of a value component.

mdelez:
> any valid `xsd:URI` (https://docs.knora.org/paradox/02-knora-ontologies/knora-base.html#subclasses-of-value, https://github.com/dasch-swiss/knora-api/blob/ccb64bbcf16b80d55423ff44b28c89d34bf39581/webapi/src/main/scala/org/knora/webapi/util/StringFormatter.scala#L713-L714).
>
I'm a little lost on how I can use this. Do we have validation functions set up already for uri's or will we still need to use regex?

> I think it would be a good idea to put the validators in a central place so they could be reused. Also then the regex would only be created once and not with every instance of a value component.
>
I can work on this. It would probably be good to implement it before we create all of the other components. Should I make a new issue for this?

projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+
+      expect(valueInputNativeElement.value).toEqual('http://www.google.ch');
+
+      expect(testHostComponent.inputValueComponent.form.valid).toBeFalsy();
+
+    });
+
+    it('should set a new display value', () => {
+
+      const newInt = new ReadUriValue();
```
tobiasschweizer: Could you adapt the variable name? Just for reasons of clarity, doesn't make any technical difference.

projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+    it('should unsubscribe when destroyed', () => {
+      expect(testHostComponent.inputValueComponent.valueChangesSubscription.closed).toBeFalsy();
+
+      testHostComponent.inputValueComponent.ngOnDestroy();
+
+      expect(testHostComponent.inputValueComponent.valueChangesSubscription.closed).toBeTruthy();
+    });
+  });
+
+  describe('create an integer value', () => {
```
tobiasschweizer: please adapt the test's name

projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
mdelez: done in b228083

projects/knora-ui/src/lib/viewer/values/uri-value/uri-value.component.spec.ts
mdelez: done in b228083

PR #23, opened 2020-02-14
Decimal Value Component

### LINKED ISSUE #8: Viewer Value Component: DecimalValueComponent
Implement `BaseValueComponent` for a `DecimalValue`.

mdelez:
and unit tests

closes #8

### DIFF
--- a/projects/knora-ui/src/lib/viewer/values/decimal-value/decimal-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/decimal-value/decimal-value.component.spec.ts
@@ -1,0 +1,339 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { DecimalValueComponent } from './decimal-value.component';
+import { ReadDecimalValue, MockResource, UpdateValue, UpdateDecimalValue, CreateDecimalValue } from '@knora/api';
+import { OnInit, Component, ViewChild, DebugElement } from '@angular/core';
+import { ReactiveFormsModule } from '@angular/forms';
+import { MatInputModule } from '@angular/material';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { $ } from 'protractor';
+import { By } from '@angular/platform-browser';
+
+/**
+ * Test host component to simulate parent component.
+ */
+@Component({
+  template: `
+    <kui-decimal-value #inputVal [displayValue]="displayInputVal" [mode]="mode"></kui-decimal-value>`
+})
+class TestHostDisplayValueComponent implements OnInit {
+
+  @ViewChild('inputVal', {static: false}) inputValueComponent: DecimalValueComponent;
+
+  displayInputVal: ReadDecimalValue;
+
+  mode: 'read' | 'update' | 'create' | 'search';
+
+  ngOnInit() {
+
+    MockResource.getTestthing().subscribe(res => {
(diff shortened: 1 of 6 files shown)


tobiasschweizer:
Great, I will try to do the review today!

projects/knora-ui/src/lib/viewer/viewer.module.ts
```diff
@@ -5,9 +5,10 @@ import {MatInputModule} from '@angular/material';
 import {CommonModule} from '@angular/common';
 import { IntValueComponent } from './values/int-value/int-value.component';
 import { DisplayEditComponent } from './operations/display-edit/display-edit.component';
+import { DecimalValueComponent } from './values/decimal-value/decimal-value.component';
 
 @NgModule({
-  declarations: [TextValueAsStringComponent, IntValueComponent, DisplayEditComponent],
+  declarations: [TextValueAsStringComponent, IntValueComponent, DisplayEditComponent, DecimalValueComponent],
   imports: [
```
tobiasschweizer: Please also add `DecimalValueComponent` to exports, otherwise it cannot be used standalone.

projects/knora-ui/src/lib/viewer/values/decimal-value/decimal-value.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  @Input() displayValue?: ReadDecimalValue;
+
+  valueFormControl: FormControl;
+  commentFormControl: FormControl;
+
+  form: FormGroup;
+
+  valueChangesSubscription: Subscription;
+
+  customValidators = [Validators.pattern(/^-?[0-9]+[.]{1}[0-9]+$/)]; // only allow decimal values
```
tobiasschweizer: You actually do not need a custom validator here: also an integer is valid decimal (but not every decimal is a valid integer). Since the template only allows for numeric values, we should be safe.

It might be worth to look at the allowed 'e' input: https://stackoverflow.com/questions/31706611/why-does-the-html-input-with-type-number-allow-the-letter-e-to-be-entered-in

mdelez: I think allowing the user to type "e" is okay. The form control seems to convert it into the actual number before it is sent to Knora accepts the value. The only thing is that maybe the user won't like that it gets converted automatically meaning they will never be able to type "2e2" and get "2e2" back. It will always return the convert value, in this case 200, because it is converted before it is sent to Knora.

projects/knora-ui/src/lib/viewer/values/decimal-value/decimal-value.component.spec.ts
```diff
@@ -0,0 +1,339 @@
...
+    it('should unsubscribe when destroyed', () => {
+      expect(testHostComponent.inputValueComponent.valueChangesSubscription.closed).toBeFalsy();
+
+      testHostComponent.inputValueComponent.ngOnDestroy();
+
+      expect(testHostComponent.inputValueComponent.valueChangesSubscription.closed).toBeTruthy();
+    });
+  });
+
+  describe('create a decimaö value', () => {
```
tobiasschweizer: typo "decimaö"

mdelez: 🤦🏻‍♂️I blame the swiss german keyboard. There are too many letters.

PR #25, opened 2020-02-17
feature (value): moved custom regex validators to a shared class

### LINKED ISSUE #24: Centralize validators
We should move the various validators into a shared centralized class so that we don't need to make a new validator every time we create a new instance of a ValueComponent.

mdelez:
closes #24

### DIFF
--- a/projects/knora-ui/src/lib/viewer/values/custom-regex.ts
+++ b/projects/knora-ui/src/lib/viewer/values/custom-regex.ts
@@ -1,0 +1,8 @@
+// @dynamic
+// https://github.com/ng-packagr/ng-packagr/issues/641
+
+export class CustomRegex{
+
+    public static readonly INT_REGEX = RegExp(/^-?\d+$/);
+    
+}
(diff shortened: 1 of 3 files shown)


projects/knora-ui/src/lib/viewer/values/custom-validators.ts
```diff
@@ -0,0 +1,5 @@
+import { Validators } from '@angular/forms';
```
tobiasschweizer: duplicate code

projects/knora-ui/src/lib/viewer/values/custom-regex.ts
```diff
@@ -0,0 +1,8 @@
+// @dynamic
+// https://github.com/ng-packagr/ng-packagr/issues/641
+
+export class CustomRegex{
+
+    public static readonly INT_REGEX = RegExp(/^-?\d+$/);
```
tobiasschweizer: I think it's fine like this `/^-?\d+$/`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp

PR #26, opened 2020-02-17
Interval Value Component

### LINKED ISSUE #13: Viewer Value Component: IntervalValueComponent
Implement `BaseValueComponent` for an `IntervalValue`.

Consider making a custom Angular Material  form field control since an interval needs two input fields but there is only one form control for the actual value: https://material.angular.io/guide/creating-a-custom-form-field-control

The actual value would be an object with a start and an end member.

tobiasschweizer:
closes #13


projects/knora-ui/src/lib/viewer/values/base-value.component.ts
```diff
@@ -53,6 +41,16 @@ export abstract class BaseValueComponent {
    */
   abstract customValidators: ValidatorFn[];
 
+  standardValidatorFunc: (val: any, comment: string, commentCtrl: FormControl) => ValidatorFn = (initValue: any, initComment: string, commentFormControl: FormControl): ValidatorFn => {
```
tobiasschweizer: @mdelez I made the standard validator a member of the base class that can be overridden if necessary.

projects/knora-ui/src/lib/viewer/values/interval-value/interval-value.component.ts
```diff
@@ -0,0 +1,115 @@
...
+  customValidators = [];
+
+  constructor(@Inject(FormBuilder) private fb: FormBuilder) {
+    super();
+  }
+
+  standardValidatorFunc: (val: any, comment: string, commentCtrl: FormControl) => ValidatorFn = (initValue: any, initComment: string, commentFormControl: FormControl): ValidatorFn => {
+    return (control: AbstractControl): { [key: string]: any } | null => {
+
+      const invalid = (initValue.start === control.value.start && initValue.end === control.value.end)
```
tobiasschweizer: @mdelez Wherever we deal with complex objects, the class needs to provide its custom validation method, overriding the standard method.

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
```diff
@@ -51,6 +51,10 @@ export class DisplayEditComponent implements OnInit {
 
   }
 
+  ngAfterViewInit(): void {
+    // console.log(this.displayValueComponent);
+  }
+
```
mdelez: Seems like this is a remnant from debugging. Can we remove it (as well as the import & export)?

src/app/app.component.ts
```diff
@@ -31,12 +31,16 @@ export class AppComponent implements OnInit {
...
+        this.testValue = this.testthing.getValues('http://0.0.0.0:3333/ontology/0001/anything/v2#hasInterval')[0];
       }
     );
 
   }
 
+  ngAfterViewInit(): void {
+    // console.log(this);
+  }
+
```
mdelez: left over debugging code that can be removed as well I think

tobiasschweizer: done in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/26/commits/59066f2762b9a001a2eefd39e6155aac23f34bd0

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
tobiasschweizer: done in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/26/commits/59066f2762b9a001a2eefd39e6155aac23f34bd0

mdelez:
Should the interval allow for a start that is a higher value than the end? Also it would be nice to have the container not "focus" and turn purple as if you are updating the value when you click on the container while the form is in read-only mode.

PR #27, opened 2020-02-17
Link value component

### LINKED ISSUE #14: Viewer Value Component: LinkValueComponent
Implement `BaseValueComponent` for a `LinkValue`.

SepidehAlassi:
_to do:_
**display:**
- [x] show the target resource label in a read-only drop-down menu

**edit:**
- [x] search for all possible resources with given three letters of the label

- [x] list the resource labels of all possible target resources in a drop-down menu

- [x] read the user's choice and assign to the target of the link value

**test**

- [x] write tests for edit and display

closes #14

### DIFF
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
@@ -18,7 +18,7 @@
     <kui-time-value #displayVal *ngSwitchCase="constants.TimeValue" [mode]="mode" [displayValue]="displayValue">
     </kui-time-value>
     <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue"
-                    [parentResource]="parentResource" [propType]="displayValue.property"></kui-link-value>
+                    [parentResource]="parentResource" [propIri]="displayValue.property"></kui-link-value>
   </span>
 </div>
 <div>
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
@@ -2,6 +2,7 @@
   <span [ngSwitch]="displayValue.type">
     <kui-text-value-as-string #displayVal *ngSwitchCase="constants.TextValue" [mode]="mode" [displayValue]="displayValue"></kui-text-value-as-string>
     <kui-int-value #displayVal *ngSwitchCase="constants.IntValue" [mode]="mode" [displayValue]="displayValue"></kui-int-value>
+    <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue"></kui-link-value>
   </span>
 </div>
 <div>
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
@@ -17,7 +17,7 @@
     </kui-interval-value>
     <kui-time-value #displayVal *ngSwitchCase="constants.TimeValue" [mode]="mode" [displayValue]="displayValue">
     </kui-time-value>
-    <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue"></kui-link-value>
+    <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue" [parentResource]="parentResource"></kui-link-value>
   </span>
 </div>
 <div>
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
@@ -17,7 +17,8 @@
     </kui-interval-value>
     <kui-time-value #displayVal *ngSwitchCase="constants.TimeValue" [mode]="mode" [displayValue]="displayValue">
     </kui-time-value>
-    <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue" [parentResource]="parentResource"></kui-link-value>
+    <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue"
+                    [parentResource]="parentResource" [propType]="displayValue.property"></kui-link-value>
   </span>
 </div>
 <div>
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
@@ -304,7 +304,8 @@ describe('DisplayEditComponent', () => {
       expect(valuesSpy.v2.values.updateValue).toHaveBeenCalledTimes(1);
 
       expect(valuesSpy.v2.values.getValue).toHaveBeenCalledTimes(1);
-      expect(valuesSpy.v2.values.getValue).toHaveBeenCalledWith(testHostComponent.readResource.id, testHostComponent.readValue.uuid);
+      expect(valuesSpy.v2.values.getValue).toHaveBeenCalledWith(testHostComponent.readResource.id,
+        testHostComponent.displayEditValueComponent.displayValue.uuid);
 
       expect(testHostComponent.displayEditValueComponent.displayValue.id).toEqual('newID');
       expect(testHostComponent.displayEditValueComponent.mode).toEqual('read');
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
@@ -39,7 +39,7 @@ class TestLinkValueComponent {
 
   @Input() displayValue;
   @Input() parentResource;
-  @Input() propType;
+  @Input() propIri;
 }
 @Component({
   selector: `kui-uri-value`,
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
@@ -39,6 +39,7 @@ class TestLinkValueComponent {
 
   @Input() displayValue;
   @Input() parentResource;
+  @Input() propType;
 }
 @Component({
   selector: `kui-uri-value`,
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
@@ -30,7 +30,16 @@ class TestTextValueAsStringComponent {
 
   @Input() displayValue;
 }
+@Component({
+  selector: `kui-link-value`,
+  template: ``
+})
+class TestLinkValueComponent {
 
+  @Input() mode;
+
+  @Input() displayValue;
+}
 @Component({
   selector: `kui-int-value`,
   template: ``
@@ -116,7 +125,8 @@ describe('DisplayEditComponent', () => {
         DisplayEditComponent,
         TestHostDisplayValueComponent,
         TestTextValueAsStringComponent,
-        TestIntValueComponent
+        TestIntValueComponent,
+        TestLinkValueComponent
       ],
       providers: [
         {
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
@@ -38,6 +38,7 @@ class TestLinkValueComponent {
   @Input() mode;
 
   @Input() displayValue;
+  @Input() parentResource;
 }
 @Component({
   selector: `kui-uri-value`,
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -2,9 +2,9 @@
   <mat-form-field class="example-full-width linkValue-field">
         <input [formControlName]="'linkValue'" type="text" placeholder="Link value" aria-label="resource" matInput [matAutocomplete]="auto" [readonly]="mode === 'read'">
 
-        <mat-autocomplete #auto="matAutocomplete">
-          <mat-option *ngFor="let option of options" [value]="option" (click) = "onOptionChange(option)">
-              {{option?.label}}
+        <mat-autocomplete #auto="matAutocomplete" [displayWith]="displayResource">
+          <mat-option *ngFor="let res of resources" [value]="res">
+              {{res?.label}}
           </mat-option>
         </mat-autocomplete>
     </mat-form-field>
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -1,5 +1,5 @@
 <span [formGroup]="form">
-  <mat-form-field class="example-full-width">
+  <mat-form-field class="example-full-width linkValue-field">
         <input [formControlName]="'linkValue'" type="text" placeholder="Link value" aria-label="resource" matInput [matAutocomplete]="auto" [readonly]="mode === 'read'">
 
         <mat-autocomplete #auto="matAutocomplete">
@@ -8,7 +8,7 @@
           </mat-option>
         </mat-autocomplete>
     </mat-form-field>
-    <mat-form-field class="large-field">
+    <mat-form-field class="large-field comment-field">
         <input matInput [formControlName]="'comment'" class="comment" placeholder="Comment" type="text" [readonly]="mode === 'read'">
     </mat-form-field>
 </span>
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -1,9 +1,9 @@
 <span [formGroup]="form">
   <mat-form-field class="example-full-width">
-        <input [formControlName]="'linkValue'" type="text" placeholder="Link value" aria-label="resource" matInput [matAutocomplete]="auto">
+        <input [formControlName]="'linkValue'" type="text" placeholder="Link value" aria-label="resource" matInput [matAutocomplete]="auto" [readonly]="mode === 'read'">
 
         <mat-autocomplete #auto="matAutocomplete">
-          <mat-option *ngFor="let option of options" [value]="option">
+          <mat-option *ngFor="let option of options" [value]="option" (click) = "onOptionChange(option)">
               {{option?.label}}
           </mat-option>
         </mat-autocomplete>
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -1,5 +1,14 @@
 <span [formGroup]="form">
-    <mat-form-field>
-      <input type="text" matInput [formControl]="valueFormControl">
+  <mat-form-field class="example-full-width">
+        <input [formControlName]="'linkValue'" type="text" placeholder="Link value" aria-label="resource" matInput [matAutocomplete]="auto">
+
+        <mat-autocomplete #auto="matAutocomplete">
+          <mat-option *ngFor="let option of options" [value]="option">
+              {{option}}
+          </mat-option>
+        </mat-autocomplete>
+    </mat-form-field>
+    <mat-form-field class="large-field">
+        <input matInput [formControlName]="'comment'" class="comment" placeholder="Comment" type="text" [readonly]="mode === 'read'">
     </mat-form-field>
 </span>
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -4,7 +4,7 @@
 
         <mat-autocomplete #auto="matAutocomplete">
           <mat-option *ngFor="let option of options" [value]="option">
-              {{option}}
+              {{option?.label}}
           </mat-option>
         </mat-autocomplete>
     </mat-form-field>
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -1,0 +1,5 @@
+<span [formGroup]="form">
+    <mat-form-field>
+      <input type="text" matInput [formControl]="valueFormControl">
+    </mat-form-field>
+</span>
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
@@ -22,7 +22,8 @@ import {of} from 'rxjs';
  */
 @Component({
   template: `
-    <kui-link-value #inputVal [displayValue]="displayInputVal" [mode]="mode" [parentResource]="parentResource" [propType]="propType"></kui-link-value>`
+    <kui-link-value #inputVal [displayValue]="displayInputVal" [mode]="mode" [parentResource]="parentResource"
+                    [propIri]="propIri"></kui-link-value>`
 })
 class TestHostDisplayValueComponent implements OnInit {
 
@@ -30,7 +31,7 @@ class TestHostDisplayValueComponent implements OnInit {
 
   displayInputVal: ReadLinkValue;
   parentResource: ReadResource;
-  propType: string;
+  propIri: string;
   mode: 'read' | 'update' | 'create' | 'search';
 
   ngOnInit() {
@@ -40,7 +41,7 @@ class TestHostDisplayValueComponent implements OnInit {
         res[0].getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasOtherThingValue', ReadLinkValue)[0];
 
       this.displayInputVal = inputVal;
-      this.propType = this.displayInputVal.property;
+      this.propIri = this.displayInputVal.property;
       this.parentResource = res[0];
       this.mode = 'read';
     });
@@ -53,19 +54,19 @@ class TestHostDisplayValueComponent implements OnInit {
  */
 @Component({
   template: `
-    <kui-link-value #inputVal [mode]="mode" [parentResource]="parentResource" [propType]="propType"></kui-link-value>`
+    <kui-link-value #inputVal [mode]="mode" [parentResource]="parentResource" [propIri]="propIri"></kui-link-value>`
 })
 class TestHostCreateValueComponent implements OnInit {
 
   @ViewChild('inputVal', {static: false}) inputValueComponent: LinkValueComponent;
   parentResource: ReadResource;
-  propType: string;
+  propIri: string;
   mode: 'read' | 'update' | 'create' | 'search';
 
   ngOnInit() {
 
     MockResource.getTestthing().subscribe(res => {
-      this.propType = 'http://0.0.0.0:3333/ontology/0001/anything/v2#hasOtherThingValue';
+      this.propIri = 'http://0.0.0.0:3333/ontology/0001/anything/v2#hasOtherThingValue';
       this.parentResource = res[0];
       this.mode = 'create';
     });
@@ -131,18 +132,18 @@ describe('LinkValueComponent', () => {
 
     it('should display an existing value', () => {
 
-      expect(testHostComponent.displayInputVal.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
-      expect(testHostComponent.displayInputVal.propertyLabel).toMatch('Another thing');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
+      expect(testHostComponent.inputValueComponent.displayValue.propertyLabel).toMatch('Another thing');
 
       expect(testHostComponent.inputValueComponent.form.valid).toBeTruthy();
 
       expect(testHostComponent.inputValueComponent.mode).toEqual('read');
 
-      expect(testHostComponent.displayInputVal.linkedResource.label).toEqual('Sierra');
-      expect(testHostComponent.displayInputVal.linkedResource.type).toEqual('http://0.0.0.0:3333/ontology/0001/anything/v2#Thing');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.label).toEqual('Sierra');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.type).toEqual('http://0.0.0.0:3333/ontology/0001/anything/v2#Thing');
       expect(valueInputNativeElement.readOnly).toEqual(true);
-
     });
+
     it('should make a link value editable', () => {
 
       testHostComponent.mode = 'update';
@@ -153,6 +154,7 @@ describe('LinkValueComponent', () => {
       expect(testHostComponent.inputValueComponent.form.valid).toBeFalsy();
 
     });
+
     it('should search for resources by their label', () => {
       const valuesSpy = TestBed.get(KnoraApiConnectionToken);
       valuesSpy.v2.search.doSearchByLabel.and.callFake(
@@ -191,6 +193,7 @@ describe('LinkValueComponent', () => {
       expect((updatedValue as UpdateLinkValue).valueHasComment).toEqual('this is a comment');
 
     });
+
     it('should return a selected resource', () => {
 
       const res = new ReadResource();
@@ -247,6 +250,7 @@ describe('LinkValueComponent', () => {
       expect(valuesSpy.v2.search.doSearchByLabel).toHaveBeenCalledWith('thing', 0, {limitToResourceClass: 'http://0.0.0.0:3333/ontology/0001/anything/v2#Thing'});
       expect(testHostComponent.inputValueComponent.resources.length).toEqual(1);
     });
+
     it('should create a value', () => {
       // simulate user input
       const res = new ReadResource();
@@ -263,8 +267,8 @@ describe('LinkValueComponent', () => {
       const newValue = testHostComponent.inputValueComponent.getNewValue();
       expect(newValue instanceof CreateLinkValue).toBeTruthy();
       expect((newValue as CreateLinkValue).linkedResourceIri).toEqual('http://rdfh.ch/0001/IwMDbs0KQsaxSRUTl2cAIQ');
-
     });
+
     it('should only create a new value if input is a resource', () => {
       // simulate user input
       expect(testHostComponent.inputValueComponent.form.valid).toBeFalsy();
@@ -279,6 +283,7 @@ describe('LinkValueComponent', () => {
 
       expect(newValue instanceof CreateLinkValue).toBeFalsy();
     });
+
     it('should reset form after cancellation', () => {
       // simulate user input
       const res = new ReadResource();
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
@@ -4,6 +4,7 @@ import { LinkValueComponent } from './link-value.component';
 import { ReadLinkValue, MockResource, UpdateValue, UpdateLinkValue, CreateLinkValue } from '@knora/api';
 import { OnInit, Component, ViewChild, DebugElement } from '@angular/core';
 import { ReactiveFormsModule } from '@angular/forms';
+import { MatInputModule } from '@angular/material';
 import { MatAutocompleteModule } from '@angular/material/autocomplete';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { $ } from 'protractor';
@@ -69,6 +70,7 @@ describe('LinkValueComponent', () => {
        ],
        imports: [
         ReactiveFormsModule,
+        MatInputModule,
         MatAutocompleteModule,
         BrowserAnimationsModule
       ],
@@ -88,7 +90,7 @@ describe('LinkValueComponent', () => {
 
     it('should display an existing value', () => {
 
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
+      expect(testHostComponent.displayInputVal.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
 
     });
   });
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
@@ -7,6 +7,7 @@ import { ReactiveFormsModule } from '@angular/forms';
 import { MatInputModule } from '@angular/material';
 import { MatAutocompleteModule } from '@angular/material/autocomplete';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import {KnoraApiConnectionToken} from '../../../core';
 import { $ } from 'protractor';
 import { By } from '@angular/platform-browser';
 
@@ -62,6 +63,11 @@ class TestHostCreateValueComponent implements OnInit {
 describe('LinkValueComponent', () => {
 
   beforeEach(async(() => {
+    const valuesSpyObj = {
+      v2: {
+        values: jasmine.createSpyObj('values', ['updateValue', 'getValue'])
+      }
+    };
     TestBed.configureTestingModule({
       declarations: [
         LinkValueComponent,
@@ -74,6 +80,12 @@ describe('LinkValueComponent', () => {
         MatAutocompleteModule,
         BrowserAnimationsModule
       ],
+      providers: [
+        {
+          provide: KnoraApiConnectionToken,
+          useValue: valuesSpyObj
+        }
+      ]
     })
     .compileComponents();
   }));
@@ -91,6 +103,14 @@ describe('LinkValueComponent', () => {
     it('should display an existing value', () => {
 
       expect(testHostComponent.displayInputVal.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
+
+      // expect(testHostComponent.inputValueComponent.form.valid).toBeTruthy();
+      //
+      // expect(testHostComponent.inputValueComponent.mode).toEqual('read');
+      //
+      // expect(valueInputNativeElement.value).toEqual('Sierra');
+      //
+      // expect(valueInputNativeElement.readOnly).toEqual(true);
 
     });
   });
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
@@ -1,0 +1,117 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { LinkValueComponent } from './link-value.component';
+import { ReadLinkValue, MockResource, UpdateValue, UpdateLinkValue, CreateLinkValue } from '@knora/api';
+import { OnInit, Component, ViewChild, DebugElement } from '@angular/core';
+import { ReactiveFormsModule } from '@angular/forms';
+import { MatInputModule } from '@angular/material';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { $ } from 'protractor';
+import { By } from '@angular/platform-browser';
+
+/**
+ * Test host component to simulate parent component.
+ */
+@Component({
+  template: `
+    <kui-link-value #inputVal [displayValue]="displayInputVal" [mode]="mode"></kui-link-value>`
+})
+class TestHostDisplayValueComponent implements OnInit {
+
+  @ViewChild('inputVal', {static: false}) inputValueComponent: LinkValueComponent;
+
+  displayInputVal: ReadLinkValue;
+
+  mode: 'read' | 'update' | 'create' | 'search';
+
+  ngOnInit() {
+
+    MockResource.getTestthing().subscribe(res => {
+      const inputVal: ReadLinkValue =
+        res[0].getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasOtherThingValue', ReadLinkValue)[0];
+
+      this.displayInputVal = inputVal;
+
+      this.mode = 'read';
+    });
+
+  }
+}
+
+/**
+ * Test host component to simulate parent component.
+ */
+@Component({
+  template: `
+    <kui-link-value #inputVal [mode]="mode"></kui-link-value>`
+})
+class TestHostCreateValueComponent implements OnInit {
+
+  @ViewChild('inputVal', {static: false}) inputValueComponent: LinkValueComponent;
+
+  mode: 'read' | 'update' | 'create' | 'search';
+
+  ngOnInit() {
+
+    this.mode = 'create';
+
+  }
+}
+
+describe('LinkValueComponent', () => {
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [
+        LinkValueComponent,
+        TestHostDisplayValueComponent,
+        TestHostCreateValueComponent
+       ],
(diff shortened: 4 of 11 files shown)


tobiasschweizer:
Thanks for this PR, I will try to review it tomorrow.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    return updatedLinkValue;
+  }
+  onOptionChange(option: ReadResource) {
+      this.setLinkValue(option);
+  }
+
+  setLinkValue(option: ReadResource) {
+    this.valueFormControl.setValue(option.label);
```
tobiasschweizer: This is not how it should be done because the flow of date is not solely controlled by `FormControl`. Two additional members `linkedResourceIRI` and `linkedResource` are required to manage the state. Maybe making the integer the reference component was not ideal since a number is primitive. It gets more complex when dealing with objects like in case of the interval value.

I see two options:

1. make the `FormControl`'s value a `ReadResource` and use autocomplete's `displayWith` input to show the label to the user (https://material.angular.io/components/autocomplete/overview#setting-separate-control-and-display-values). Then get rid of the additional class members, you can access the `ReadResource` via the `FormControl`. See https://github.com/dasch-swiss/knora-ui/blob/master/projects/knora/search/src/lib/extended-search/select-property/specify-property-value/link-value/link-value.component.html.

2. if 1. does not do the job, you can always make a custom subcomponent as we did for the interval or color value. However, I think 1. is preferable since it seems simpler to me.

If you need any help, please let me know.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
```diff
@@ -0,0 +1,14 @@
+<span [formGroup]="form">
+  <mat-form-field class="example-full-width linkValue-field">
+        <input [formControlName]="'linkValue'" type="text" placeholder="Link value" aria-label="resource" matInput [matAutocomplete]="auto" [readonly]="mode === 'read'">
+
+        <mat-autocomplete #auto="matAutocomplete">
+          <mat-option *ngFor="let option of options" [value]="option" (click) = "onOptionChange(option)">
```
tobiasschweizer: I think if the `FormControl` contains a `ReadResource`, then `(click) = "onOptionChange(option)"` becomes obsolete. In `getNewValue` and `getUpdatedValue` you can then directly access the `FormControl`'s value and get the IRI of the `ReadResource`.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+          res.id = 'http://rdfh.ch/0001/IwMDbs0KQsaxSRUTl2cAIQ';
+          res.label = 'hidden thing';
+          return of([res]);
+        }
+      );
+
+      testHostComponent.inputValueComponent.searchByLabel('thing');
+
+      testHostFixture.detectChanges();
+
```
tobiasschweizer: I think you should add additional assertions like:

`expect(valuesSpy.v2.search.doSearchByLabel).toHaveBeenCalledWith('thing', 0, { limitToResourceClass: 'http://0.0.0.0:3333/ontology/0001/anything/v2#Thing'});`

This makes sure that the method has been called with the correct args.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,151 @@
...
+  getInitValue(): ReadResource | null {
+    if (this.displayValue !== undefined) {
+      return this.displayValue.linkedResource;
+    } else {
+      return null;
+    }
+  }
+  // override the resetFormControl() from the base component to deal with initial link value label
+  resetFormControl(): void {
+    super.resetFormControl();
```
tobiasschweizer: I think it is enough to do:

```ts
  // override the resetFormControl() from the base component to deal with initial link value label
  resetFormControl(): void {
    super.resetFormControl();

    if (this.valueFormControl !== undefined) {
      if (this.mode !== 'read') {

        this.valueChangesSubscription = this.valueFormControl.valueChanges.subscribe(data => {
          this.options = this.searchByLabel(data);
        });

      }
    }
  }
```

Please also unsubscribe from this in `ngOnDestroy` (like we do for the comment change subscription).

Perhaps you also need to include this here from `ngOnInit`:

`this.options = [this.displayValue.linkedResource];`

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,151 @@
...
+  linkedResourceIRI: string;
+
+  valueChangesSubscription: Subscription;
+  // label cannot contain logical operations of lucene index
+  customValidators = [];
+
+  constructor(@Inject(FormBuilder) private fb: FormBuilder, @Inject(KnoraApiConnectionToken) private knoraApiConnection: KnoraApiConnection) {
+    super();
+  }
+
```
tobiasschweizer: I think since the value is a an object (`ReadResource`), you need to override the standard validator function:

```ts
  standardValidatorFunc: (val: any, comment: string, commentCtrl: FormControl) => ValidatorFn = (initValue: any, initComment: string, commentFormControl: FormControl): ValidatorFn => {
    return (control: AbstractControl): { [key: string]: any } | null => {

      const invalid = (!(control.value instanceof ReadResource) || initValue.id === control.value.id)
        && (initComment === commentFormControl.value || (initComment === null && commentFormControl.value === ''));

      return invalid ? {valueNotChanged: {value: control.value}} : null;
    };
  };
```

tobiasschweizer: In order to be valid, the value has to be a `ReadResource` (value is a string while the user types) and the id has to be different from the initial resource.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+    it('should return a selected resource', () => {
+
+      const res = new ReadResource();
+      res.id = 'http://rdfh.ch/0001/a-blue-thing';
+      testHostComponent.inputValueComponent.setLinkValue(res);
+
+      testHostFixture.detectChanges();
+
+      expect(testHostComponent.inputValueComponent.linkedResourceIRI).toEqual('http://rdfh.ch/0001/a-blue-thing');
+    });
```
tobiasschweizer: Could you please add tests for the value creation case?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Search for resources whose labels contain the given search term, restricting to to the given properties object constraint.
+   * this is to be used for update and new linked resources
+   *
+   * @param label to be searched
+   */
+    searchByLabel(searchTerm: string): ReadResource[] {
+      this.restrictToResourceClass = this.displayValue.linkedResource.type;
```
tobiasschweizer: Could you make this a local variable?

What about the value creation case? I think you need to look at the ontology for the property in question to find about about the restriction.

tobiasschweizer: Probably the component needs an additional input where the property definition can be passed in.

tobiasschweizer:
@SepidehAlassi Please let me know if you have time to work on this, otherwise I will be glad to help out.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
SepidehAlassi: yes, just doing that.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.html
SepidehAlassi: done!

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: thanks for this, option 1 works.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: @tobiasschweizer yes, that is right. How can I check the object type of a link property from Ontology? Is there a function in @knora/Api for returning property definition?

tobiasschweizer: You can get it from the `entityDefinition` on the resource that you are editing / creating a value for. I think you need an additional Input for your component where you pass this information in. I can have a look and make an example if this helps.

tobiasschweizer: Here is the member on `ReadResource`:

https://github.com/dasch-swiss/knora-api-js-lib/blob/03e74aec56d7d005b866e6f48ddf6b771fa280a0/src/models/v2/resources/read/read-resource.ts#L45

From that collection, the information for the property in question would have to be returned. 

That would be https://github.com/dasch-swiss/knora-api-js-lib/blob/03e74aec56d7d005b866e6f48ddf6b771fa280a0/src/models/v2/ontologies/property-definition.ts#L9

The `objectType` is optional. So when absent, the restriction is a `knora-api:Resource`

tobiasschweizer: for the edit case, you can get it from here

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/e9cae9fa155c22a02385dc3a6558243525d7b3a1/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts#L28

and pass it along in the template.

SepidehAlassi: Ok, I got the property definition from ontology and passed it as input to the component.
passing the parameter as below in displayEdit
` <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue" [propertyDesc]="parentResource.entityInfo"></kui-link-value>
    `
and reading it as input in my component
`  @Input() propertyDesc?: IResourceClassAndPropertyDefinitions;
console.log(this.propertyDesc.properties[this.displayValue.property])
`
See the output of the console command given above here:
image

Keeping in mind how we define the link properties in ontology: we specify the type of the object resource in property definition itself e.g. "hasOtherThing" but the "hasOtherThingValue" is actually the LinkValue. In app.component, I read "hasOtherThingValue" to set the value of the link. (i.e. this.displayValue.property is http://0.0.0.0:3333/ontology/0001/anything/v2#hasOtherThingValue)

The question is how to read the objectType of the "hasOtherThing"? Regex is not an option, one cannot simply trim "Value" from the end of this.displayValue.property.

I believe there should be a way to get the linking property ("hasOtherThing") definition from the definition of its linkValue ("hasOtherThingValue").

tobiasschweizer: > I believe there should be a way to get the linking property ("hasOtherThing") definition from the definition of its linkValue ("hasOtherThingValue").

True, good point! I could make a function in knora-api-js-lib. 
In Knora, there is the same logic.

tobiasschweizer: Ok, I will try to do this and let you know!

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
SepidehAlassi: added!

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: thanks, added this.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: well, in resetFromControl, I need to set the label of the linkedResource initial value of the link value as well.

tobiasschweizer: Sorry, I don't understand. In `getInitValue` you are returning the linked resource from the link value. Hence this `ReadResource` is set for the `FormControl`'s value (default behavior of `resetFormControl`). Since there is `[displayWith]="displayResource"`, the component knows what to display in the input field (the resource's label).

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,175 @@
...
+    return this.resources;
+  }
+
+  // show the label of the linked resource
+  getInitValue(): ReadResource | null {
+    if (this.displayValue !== undefined) {
+
+      // console.log(this.displayValue);
+      // console.log(this.propertyDesc.properties[this.displayValue.property]);
+      return this.displayValue.linkedResource;
```
tobiasschweizer: Actually this member is optional. In this context it is save to assume that the linked resource is present. However, when dealing with Gravsearch results it is not (a link value might only contain the Iri of the linked resource but no embedded object to prevent circular references).

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: Please check https://github.com/dasch-swiss/knora-api-js-lib/pull/163

tobiasschweizer:
Would it be possible to finish this PR before I do the review of the list value PR? I am trying to limit parallel tasks.

SepidehAlassi:
> Would it be possible to finish this PR before I do the review of the list value PR? I am trying to limit parallel tasks.

I believe this one is now ready to be reviewed. Please let me know if I have missed addressing any of your points.

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
```diff
@@ -29,7 +29,18 @@ class TestTextValueAsStringComponent {
...
+  selector: `kui-link-value`,
+  template: ``
+})
+class TestLinkValueComponent {
 
+  @Input() mode;
+
+  @Input() displayValue;
+  @Input() parentResource;
+  @Input() propType;
```
SepidehAlassi: @tobiasschweizer I have added the property type as an input parameter for the create mode. In displayEdit mode, it is set to displayValue.property. However, for the creation mode, I needed to know the property type. Is there any better way for this?

tobiasschweizer: Sorry, I am not sure I understand correctly. Couldn't you use the same Input prop for both cases, edit and create?

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
```diff
@@ -16,6 +17,8 @@
     </kui-interval-value>
     <kui-time-value #displayVal *ngSwitchCase="constants.TimeValue" [mode]="mode" [displayValue]="displayValue">
     </kui-time-value>
+    <kui-link-value #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue"
```
tobiasschweizer: My idea was to have an additional input with the type restriction of the linked resource. I think you can get that type from the definition, there is no need for the current value to be involved. Then it also works for create,

tobiasschweizer: you just need to know about the property, you don't need an existing value.

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
SepidehAlassi: @propType is an input parameter for both display/edit and create mode. 
If the mode === display or update: 
then propType=displayValue.property
in the create mode, proptype parameter should be specified as a parameter. See the template of component for both cases.

tobiasschweizer: so `propType` is the IRI of the property and there is another Input for the parent resource holding the entity defs?

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
SepidehAlassi: @tobiasschweizer to restrict the linked resource to the ObjectType of the link property, I need to know the properties of the parent resource. that is why the parent resource should be passed as the parameter.

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
SepidehAlassi: yes

tobiasschweizer: so there needs to be some logic that gets the prop definition from the entities

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
tobiasschweizer: see https://github.com/dasch-swiss/knora-ui-ng-lib/pull/27#discussion_r393775872

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
tobiasschweizer: if you want, I can have a look at it and see how the logic could look like

SepidehAlassi: @tobiasschweizer well, I do not see any problem with the current implementation. it works for both display/edit and the create mode.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Search for resources whose labels contain the given search term, restricting to to the given properties object constraint.
+   * this is to be used for update and new linked resources
+   *
+   * @param label to be searched
+   */
+  searchByLabel(searchTerm: string): ReadResource[] {
+    const linkType = this.parentResource.getLinkPropertyIriFromLinkValuePropertyIri(this.propType);
+    if (typeof linkType === 'string') {
+      this.restrictToResourceClass = this.parentResource.entityInfo.properties[linkType].objectType;
```
tobiasschweizer: I think object type is optional. If not defined, you can omit the restriction.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * Search for resources whose labels contain the given search term, restricting to to the given properties object constraint.
+   * this is to be used for update and new linked resources
+   *
+   * @param label to be searched
+   */
+  searchByLabel(searchTerm: string): ReadResource[] {
+    const linkType = this.parentResource.getLinkPropertyIriFromLinkValuePropertyIri(this.propType);
+    if (typeof linkType === 'string') {
```
tobiasschweizer: Why is that needed?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /**
+   * Search for resources whose labels contain the given search term, restricting to to the given properties object constraint.
+   * this is to be used for update and new linked resources
+   *
+   * @param label to be searched
+   */
+  searchByLabel(searchTerm: string): ReadResource[] {
+    const linkType = this.parentResource.getLinkPropertyIriFromLinkValuePropertyIri(this.propType);
```
tobiasschweizer: Could this be evaluated only once?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+        && (initComment === commentFormControl.value || (initComment === null && commentFormControl.value === ''));
+
+      return invalid ? {valueNotChanged: {value: control.value}} : null;
+    };
+  };
+
+  // override the resetFormControl() from the base component to deal with initial link value label
+  resetFormControl(): void {
+    super.resetFormControl();
+
```
tobiasschweizer: I think it is enough to do

```
this.valueChangesSubscription = this.valueFormControl.valueChanges.subscribe(data => {
           this.resources = this.searchByLabel(data);
         });
```

Also please unsubscribe on component destruction.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (this.valueFormControl !== undefined) {
+      if (this.mode !== 'read') {
+        this.valueFormControl.setValue('');
```
tobiasschweizer: I think this should be removed because the user probably just wants to change the comment.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+        this.valueChangesSubscription = this.valueFormControl.valueChanges.subscribe(data => {
+          this.resources = this.searchByLabel(data);
+        });
+        this.valueFormControl.clearValidators();
```
tobiasschweizer: This removes all validators, `standardValidatorFunc` will have no effect.

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
tobiasschweizer: Sorry, I missed this at first. It is now clear to me.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: because getLinkPropertyFromLinkValueIRI returns string || Error.

SepidehAlassi: without this check, the compiler complains about setting Error as index.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: hmm, yes. I can move it to ngOnInit

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: @tobiasschweizer I do unsubscribe on destroy.
`ngOnDestroy(): void {
    this.unsubscribeFromValueChanges();
  }`
is it not enough?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: if the user only changes comment, the initial value is set back. See 
` // add IRI to updatedLinkValue only if the user submitted a new Link value
    if (this.valueFormControl.value !== null && this.valueFormControl.value !== '') {
      updatedLinkValue.linkedResourceIri = this.valueFormControl.value.id;
    } else {
      // if user just updated comment value keep the old link value
      updatedLinkValue.linkedResourceIri = this.displayValue.linkedResourceIri;
    }`

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: I thought objectType should be definitely specified for link properties.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: oops, that's a bug  in the lib, will fix it

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: According to [documentation](https://docs.knora.org/paradox/02-knora-ontologies/knora-base.html#constraints-on-the-types-of-property-subjects-and-objects), subjectClassConstraing is optional for linkValues but ObjectClassConstraint is mandatory. 
@benjamingeer I remember this from the time I was writing ontologies. Is my understanding still correct?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: fixed in https://github.com/dasch-swiss/knora-api-js-lib/pull/168

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: > According to [documentation](https://docs.knora.org/paradox/02-knora-ontologies/knora-base.html#constraints-on-the-types-of-property-subjects-and-objects), subjectClassConstraing is optional for linkValues but ObjectClassConstraint is mandatory.

I think `objectType` optional in `ResourcePropertyDefinition` because there are properties that do not have an object class constraint (`ResourcePropertyDefinition` has to represent any property). Because of that, it's optional for all properties in `ResourcePropertyDefinition`.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: I fixed it, I think you can safely remove that check.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: I think that would do it, since the object type of the property does not change (unless of course there are changes in the ontology). 

Could you put that logic in a method and write a test for it?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: The name is misleading, this is for a changes subscription for the comment. We should change that name later (see #45 ).

```ts
this.labelChangesSubscription = this.valueFormControl.valueChanges.subscribe(data => {
           this.resources = this.searchByLabel(data);
         });
```

and then in `ngOnDestroy` add:

```ts
if (this.labelChangesSubscription !== undefined) {
    this.labelChangesSubscription.unsubscribe();
}
```

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: Since the existing value is initialised, couldn't you just leave it until the user decides to change it (or not)?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+   *
+   * @param label to be searched
+   */
+  searchByLabel(searchTerm: string): ReadResource[] {
+    const linkType = this.parentResource.getLinkPropertyIriFromLinkValuePropertyIri(this.propType);
+    if (typeof linkType === 'string') {
+      this.restrictToResourceClass = this.parentResource.entityInfo.properties[linkType].objectType;
+    }
+    // at least 3 characters are required
+    if (searchTerm.length >= 3) {
```
tobiasschweizer: Please check that `searchTerm` is a string: `typeof searchTerm === 'string' && searchTerm.length >= 3`

Now I understand why you set the initial value to an empty string. With the additional check above the standard behaviour should be fine. With the standard behaviour the form value could be: `null`, `ReadResource`, or a `string`.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: Please move 

```ts
this.valueFormControl.valueChanges.subscribe(data => {
           this.resources = this.searchByLabel(data);
         });
```

 to `ngOnInit`

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
benjamingeer: Which resource properties have no object class constraint? Maybe I’m forgetting something, but I think there shouldn’t be any.

tobiasschweizer: I think it is the `resourceProperty`:

```
###  http://www.knora.org/ontology/knora-base#resourceProperty

:resourceProperty rdf:type owl:ObjectProperty ;

          rdfs:comment "The base property of properties that point from Knora resources to Knora resources or values. These properties are required to have cardinalities in the resource classes in which they are used."@en ;

          :subjectClassConstraint :Resource .
```

tobiasschweizer: and then in the v2 ontology:

```json
{
    "@id" : "knora-api:resourceProperty",
    "@type" : "owl:ObjectProperty",
    "knora-api:isResourceProperty" : true,
    "knora-api:subjectType" : {
      "@id" : "knora-api:Resource"
    },
    "rdfs:comment" : "The base property of properties that point from Knora resources to Knora resources or values. These properties are required to have cardinalities in the resource classes in which they are used."
  }
```

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: see also https://github.com/dasch-swiss/knora-ui-ng-lib/pull/27#pullrequestreview-376625522

tobiasschweizer:
please see my proposal #47

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,169 @@
...
+  /**
+   * Displays a selected resource using its label.
+   *
+   * @param resource the resource to be displayed (or no selection yet).
+   * @returns
+   */
+  displayResource(resource: ReadResource | null): string {
+
+    // null is the initial value (no selection yet)
+    if (resource !== null) {
```
tobiasschweizer: Maybe it would be even better to check for `instance of `ReadResource``

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
benjamingeer: But `knora-api:resourceProperty` can never be used in a resource. It's only an abstract base property. Would it help if we changed it to have `"knora-api:isResourceProperty" : false`?

tobiasschweizer: The thing is that I have to parse it and turn into a class. And if that class had a non optional objectType, it would fail.

benjamingeer: You said:

> `objectType` optional in `ResourcePropertyDefinition` because there are properties that do not have an object class constraint

I think this doesn't make sense. All resource properties have an `objectType`. If the only exception is `knora-api:resourceProperty`, which is not really a resource property but only the base property of all resource properties, we can change `knora-api:resourceProperty` so it's not considered a resource property. Then you can make `objectType` required instead of optional in `ResourcePropertyDefinition`.

tobiasschweizer: So then we would not turn `knora-api:isResourceProperty` into a `ResourcePropertyDefinition`?

benjamingeer: > So then we would not turn knora-api:isResourceProperty into a ResourcePropertyDefinition?

Do you mean `knora-api:resourceProperty` or `knora-api:isResourceProperty`?

tobiasschweizer: > > So then we would not turn knora-api:isResourceProperty into a ResourcePropertyDefinition?
> 
> Do you mean `knora-api:resourceProperty` or `knora-api:isResourceProperty`?

I mean `knora-api:resourceProperty`.

If you set `knora-api:isResourceProperty` to `false` for `knora-api:resourceProperty`, then it won't be converted into a `ResourcePropertyDefinition` (I think then it will be just ignored).

https://github.com/dasch-swiss/knora-api-js-lib/blob/a99e32d0c6f7b8f12e2a952716c5270d8999e9f0/src/models/v2/ontologies/OntologyConversionUtil.ts#L73-L81

I suggest to create issues for this in knora-api and knora-api-js-lib and do it when we have time for it.

tobiasschweizer: Or should it be handled like a system property?

benjamingeer: > If you set knora-api:isResourceProperty to false for knora-api:resourceProperty, then it won't be converted into a ResourcePropertyDefinition (I think then it will be just ignored).

Yes, I think that would make sense. The client code should always be able to trust that a resource property will have an object type.

> I suggest to create issues for this in knora-api and knora-api-js-lib and do it when we have time for it.

OK, will do.

tobiasschweizer: system properties are expected to to have the `knora-api:isResourceProperty` flag:

https://github.com/dasch-swiss/knora-api-js-lib/blob/a99e32d0c6f7b8f12e2a952716c5270d8999e9f0/src/models/v2/ontologies/OntologyConversionUtil.ts#L83-L91

benjamingeer: > Or should it be handled like a system property?

It's never used in resources. It's only there as documentation. So I think client code will never use it.

tobiasschweizer: If I remember correctly, `knora-api:isResourceProperty` is absent or present. If present, it is set to `true`. Or am I mistaken?

tobiasschweizer: absence is like implicitly `false`

benjamingeer: > system properties are expected to not have the knora-api:isResourceProperty flag:

That makes sense. System properties are not resource properties. The idea of a resource property is that its object is a `knora-api:Value` or a `knora-api:Resource`.

> If I remember correctly, knora-api:isResourceProperty is absent or present. If present, it is set to true.

Yes.

> absence is like implicitly false

Yes.

tobiasschweizer:
Shall I take another look?

SepidehAlassi:
> Shall I take another look?

if you wish so

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: I think with https://github.com/dasch-swiss/knora-api-js-lib/pull/168 having been merged, this check is not necessary anymore.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -0,0 +1,312 @@
...
+      valueInputNativeElement = valueComponentDe.query(By.css('input')).nativeElement;
+
+      commentInputDebugElement = valueComponentDe.query(By.css('input.comment'));
+      commentInputNativeElement = commentInputDebugElement.nativeElement;
+
+    });
+
+    it('should display an existing value', () => {
+
+      expect(testHostComponent.displayInputVal.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
```
tobiasschweizer: Was your intention to check for `inputValueComponent.displayValue`?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,152 @@
...
+  resources: ReadResource[];
+  restrictToResourceClass: string;
+  valueFormControl: FormControl;
+  commentFormControl: FormControl;
+  form: FormGroup;
+
+  valueChangesSubscription: Subscription;
+  labelChangesSubscription: Subscription;
+  // label cannot contain logical operations of lucene index
+  customValidators = [];
```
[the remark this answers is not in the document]
tobiasschweizer: the use case is here is more complex because the value of the `FormControl` is a string / label as long as the user types and an instance of `ReadResource` once the user selects from the list.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
SepidehAlassi: the intention was to check the iri of the linkedresource.

tobiasschweizer: you mean on the test component?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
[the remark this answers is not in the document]
SepidehAlassi: > the use case is here is more complex because the value of the `FormControl` is a string / label as long as the user types and an instance of `ReadResource` once the user selects from the list.

If I understand what you suggest correctly, you mean there should be a validator to check that a chosen option is an instance of ReadResource. However, in "update", "create", and "search" modes,  the dropdown menu shows the options for the user to choose from. These options are always instances of ReadResource. The user does not have any other choice than choosing one item from the menu. Hence, an option selected by the user is always an instance of ReadResource. That would mean that such a double validation would not actually be necessary.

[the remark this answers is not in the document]
tobiasschweizer: > However, in "update", "create", and "search" modes, the dropdown menu shows the options for the user to choose from. These options are always instances of ReadResource. The user does not have any other choice than choosing one item from the menu. Hence, an option selected by the user is always an instance of ReadResource. That would mean that such a double validation would not actually be necessary.

Yes, but as long as the user types, the `FormControl`'s value is just a string (the label). So this is to make sure that the user selected an option from the list.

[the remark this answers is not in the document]
SepidehAlassi: the "save" button is inactive unless an option is chosen from the dropdown menu.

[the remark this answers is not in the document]
tobiasschweizer: this is true for update mode because the check is integrated into standardValidatorFunc

please try with create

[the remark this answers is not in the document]
tobiasschweizer: the standard behavior is not enough here since `Validators.required` checks that the value is not `null` (https://angular.io/api/forms/Validators#required). In the case to the link value, a string is not valid, only a `ReadResource`.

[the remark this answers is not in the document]
SepidehAlassi: do we have already a component for create mode?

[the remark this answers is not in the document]
tobiasschweizer: no, not yet. But you test it by directly adding link value component to `app.component.html`. I think @mdelez will wok on a create component soon.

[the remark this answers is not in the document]
tobiasschweizer: or you can just try it out in your spec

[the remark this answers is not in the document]
SepidehAlassi: I am trying out in spec for a while now. It would have been nice to have the component.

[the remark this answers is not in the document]
tobiasschweizer: yes! Now we can make it a user story :-)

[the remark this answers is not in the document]
mdelez: @SepidehAlassi I can start to work on the create component tomorrow

SepidehAlassi:
@tobiasschweizer I have added the missing validator.

SepidehAlassi:
@tobiasschweizer I think this PR can really be merged, oder?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,160 @@
...
+@Component({
+  selector: 'kui-link-value',
+  templateUrl: './link-value.component.html',
+  styleUrls: ['./link-value.component.scss']
+})
+
+export class LinkValueComponent extends BaseValueComponent implements OnInit, OnChanges, OnDestroy {
+  @Input() displayValue?: ReadLinkValue;
+  @Input() parentResource: ReadResource;
+  @Input() propType: string;
```
tobiasschweizer: Could you rename this to property? It's just a property IRI, right?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,160 @@
...
+  resources: ReadResource[];
```
tobiasschweizer: I think you could init this with an empty array.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: > hmm, yes. I can move it to ngOnInit

Good idea!

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: I think you can remove the check for string.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
tobiasschweizer: I think you are checking the value on the mocked parent, not on the component that ought to be tested here.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    // at least 3 characters are required
+    if ( typeof searchTerm === 'string' && searchTerm.length >= 3) {
+
+      this.knoraApiConnection.v2.search.doSearchByLabel(searchTerm, 0, {limitToResourceClass: this.restrictToResourceClass}).subscribe(
+        (response: ReadResource[]) => {
+          this.resources = response;
+        });
+    } else {
+      // clear selection
+      this.resources = undefined;
```
tobiasschweizer: If think you could assign an empty array (see my comment above).

tobiasschweizer:
> @tobiasschweizer I think this PR can really be merged, oder?

I think you are almost done, please see my comments.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+
+      expect(testHostComponent.inputValueComponent.form.valid).toBeTruthy();
+
+      expect(testHostComponent.inputValueComponent.mode).toEqual('read');
+
+      expect(testHostComponent.displayInputVal.linkedResource.label).toEqual('Sierra');
+      expect(testHostComponent.displayInputVal.linkedResource.type).toEqual('http://0.0.0.0:3333/ontology/0001/anything/v2#Thing');
+      expect(valueInputNativeElement.readOnly).toEqual(true);
+
+    });
```
tobiasschweizer: nitpicking: please add a newline between the `it`s

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -0,0 +1,311 @@
...
+    it('should only create a new value if input is a resource', () => {
+      // simulate user input
+      expect(testHostComponent.inputValueComponent.form.valid).toBeFalsy();
+
+      const res = 'thing';
+      testHostComponent.inputValueComponent.valueFormControl.setValue(res);
+
+      expect(testHostComponent.inputValueComponent.valueFormControl.value instanceof ReadResource).toBeFalsy();
+      expect(testHostComponent.inputValueComponent.form.valid).toBeFalsy();
```
tobiasschweizer: Nice! So we are sure that the validator works.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: Even though I believe propType was clear, I changed it to propIri.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: done!

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: Then probably we need to agree on a consistent terminology. Shall I make then issue for that so we can discuss that independently of this PR?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
SepidehAlassi: I have already moved the call to this function into ngOnInit.

tobiasschweizer: but

```const linkType = this.parentResource.getLinkPropertyIriFromLinkValuePropertyIri(this.propType);
this.restrictToResourceClass = this.parentResource.entityInfo.properties[linkType].objectType;```

is still in `searchByLabel `. Sorry, I am confused now.

Comment (tobiasschweizer) [reply projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts]: what I meant is: instead of re-evaluating the link property's object type for each search, it could be just done once.

Comment (SepidehAlassi) [reply projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts]: Ah, I see. The confusion was due to the previous discussion where we discussed the search for label should be in OnInit not in resetFormControl. I just moved the evaluation of the this.restrictResourceClass to the ngOnInit.

**Comment by SepidehAlassi** [comment]
@tobiasschweizer Can I merge this now?

File: projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts [hunk 26]
```diff
@@ -57,21 +57,15 @@ export class LinkValueComponent extends BaseValueComponent implements OnInit, On
...
     // at least 3 characters are required
     if ( typeof searchTerm === 'string' && searchTerm.length >= 3) {
 
       this.knoraApiConnection.v2.search.doSearchByLabel(searchTerm, 0, {limitToResourceClass: this.restrictToResourceClass}).subscribe(
         (response: ReadResource[]) => {
           this.resources = response;
         });
-    } else {
-      // clear selection
-      this.resources = undefined;
```
tobiasschweizer: I think the idea of clearing the selection is good, what I meant is just `this.resources = []`

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -0,0 +1,154 @@
...
+  selector: 'kui-link-value',
+  templateUrl: './link-value.component.html',
+  styleUrls: ['./link-value.component.scss']
+})
+
+export class LinkValueComponent extends BaseValueComponent implements OnInit, OnChanges, OnDestroy {
+  @Input() displayValue?: ReadLinkValue;
+  @Input() parentResource: ReadResource;
+  @Input() propIri: string;
+  resources: ReadResource[] = [];
```
SepidehAlassi: @tobiasschweizer I have already initialized it here to an empty array

SepidehAlassi:
> looks good, thanks
> 
> I have just one last comment: [#27 (review)](https://github.com/dasch-swiss/knora-ui-ng-lib/pull/27#pullrequestreview-381177939)

see

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.ts
tobiasschweizer: I understood that your idea of clearing selection empties the list when the user types and the string is still too short. I think this idea is good.

tobiasschweizer: I think this is important when user selects a resource and then wants to change it again. Then the selection has to change to.

tobiasschweizer: because I think otherwise you still get the old selection when the label string is shorter than 3 chars

tobiasschweizer: but maybe I am wrong, I will leave this to you. Please proceed as you want.

PR #28, opened 2020-02-17
Automatically look for deactivated tests

tobiasschweizer:
Automatically searches for deactivated tests.


tobiasschweizer:
As a next step, we should run a linter.

PR #30, opened 2020-02-19
Mock KnoraApiConnection in Tests

tobiasschweizer:
I think I can come up with an alternative to using `inject` in an `it`.

Turns out that `KnoraApiCnnection` can be mocked with a simple object that defines spies for the methods that are actually called by the testes code.


tobiasschweizer:
@flavens I think this is a better approach since it is quite consistent with the Angular testing guide:  https://v8.angular.io/guide/testing#angular-testbed

Issue #4, opened 2020-02-13
Value components


tobiasschweizer:
- `TextValue`:
    - [x]  `TextValueAsStringComponent`
    - [ ] `TextValueAsXMLComponent` (#5)
    - [ ] `TextValueAsHTMLComponent` (readonly, no editing) (#6)
- [ ] `IntValueComponent` (tests still have to be written) (#7) -> serves as reference implementation
- [ ] `DecimalValueComponent` (#8)
- [ ] `BooleanValueComponent` (#9)
- [ ] `ColorValueComponent` (#10)
- [ ] `DateValueComponent` (#11)
- [ ] `IntervalValueComponent` (#13)
- [ ] `LinkValueComponent` (#14)
- [ ] `ListValueComponent` (#15)
- [ ] `TimeValueComponent` (#16)
- [ ] `UriValueComponent` (#17)

To be defined:
- [ ] `FileValueComponent` (image, text, video, DDD)
- [ ] `GeometryComponent`
- [ ] `GeonameValueComponent` (#12)

tobiasschweizer:
A value component has to implement `BaseValueComponent`. We will use `IntValueComponent` as an example.

- create a new value component: `ng g c viewer/values/IntValue --project knora-ui --prefix=kui`
- extend the base class, implement interfaces: `class IntValueComponent extends BaseValueComponent implements OnInit, OnChanges, OnDestroy`
- implement all abstract members from the base class: `displayValue`, `valueFormControl`, `commentFormControl`, `form`, `valueChangesSubscription`, `customValidators`, see https://github.com/dasch-swiss/knora-ui-ng-lib/blob/dfd3f514883de0c76961c1768f863d5dbb1dbd3a/projects/knora-ui/src/lib/viewer/values/int-value/int-value.component.ts#L14-L23
- inject `FormBuilder` and call the base class's constructor in the value component's constructor, see https://github.com/dasch-swiss/knora-ui-ng-lib/blob/dfd3f514883de0c76961c1768f863d5dbb1dbd3a/projects/knora-ui/src/lib/viewer/values/int-value/int-value.component.ts#L25-L27
- implement `getInitValue` for the given value, see https://github.com/dasch-swiss/knora-ui-ng-lib/blob/dfd3f514883de0c76961c1768f863d5dbb1dbd3a/projects/knora-ui/src/lib/viewer/values/int-value/int-value.component.ts#L29-L35
[...]

PR #34, opened 2020-02-24
Color value component

### LINKED ISSUE #10: Viewer Value Component: ColorValueComponent
Implement `BaseValueComponent` for a `ColorValue`.

flavens:
closes #10

### DIFF
--- a/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
@@ -1,10 +1,11 @@
 <div [formGroup]="colorForm" class="color-picker-container">
     <mat-form-field appearance="outline" [floatLabel]="'always'">
         <mat-label>{{ colorLabel }}</mat-label>
+        <!-- check the ngx-color-picker doc to know more about the options -->
         <input matInput class="color-picker-input color" [formControlName]="'color'" [placeholder]="colorLabel"
             [(colorPicker)]="colorVal" [style.background]="colorVal" [style.color]="colorVal"
-            (colorPickerChange)="_handleInput(colorVal)" [cpOutputFormat]="'hex'" [cpDisabled]="readonly"
-            [cpDisableInput]="true" />
+            (colorPickerChange)="_handleInput(colorVal)" [cpOutputFormat]="'hex'" [cpFallbackColor]="'#ff0000'"
+            [cpDisabled]="readonly" readonly />
         <mat-error *ngIf="colorForm.controls.color.hasError('required')">
             Color is <strong>required</strong>
         </mat-error>
--- a/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
@@ -1,0 +1,7 @@
+<mat-form-field appearance="outline" [floatLabel]="'always'">
+    <!-- <mat-label>{{ colorLabel }}</mat-label> -->
+    <input matInput [formControlName]="'color'" [placeholder]="colorLabel" [(colorPicker)]="value.color"
+        (colorPickerChange)="_handleInput()" [cpOutputFormat]="'hex'" />
+</mat-form-field>
+
+<!-- [style.background]="colorInput" [style.color]="colorInput" -->
--- a/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
@@ -1,8 +1,12 @@
 <div [formGroup]="colorForm" class="color-picker-container">
     <mat-form-field appearance="outline" [floatLabel]="'always'">
         <mat-label>{{ colorLabel }}</mat-label>
-        <input matInput [formControlName]="'color'" [placeholder]="colorLabel" [(colorPicker)]="colorVal"
-            [style.background]="colorVal" [style.color]="colorVal" (colorPickerChange)="_handleInput(colorVal)"
-            [cpOutputFormat]="'hex'" [cpDisabled]="readonly" [cpDisableInput]="true" />
+        <input matInput class="color-picker-input color" [formControlName]="'color'" [placeholder]="colorLabel"
+            [(colorPicker)]="colorVal" [style.background]="colorVal" [style.color]="colorVal"
+            (colorPickerChange)="_handleInput(colorVal)" [cpOutputFormat]="'hex'" [cpDisabled]="readonly"
+            [cpDisableInput]="true" />
+        <mat-error *ngIf="colorForm.controls.color.hasError('required')">
+            Color is <strong>required</strong>
+        </mat-error>
     </mat-form-field>
 </div>
--- a/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
@@ -1,10 +1,10 @@
 <div [formGroup]="colorForm" class="color-picker-container">
     <mat-form-field appearance="outline" [floatLabel]="'always'">
-        <mat-label>{{ colorVal }}</mat-label>
+        <mat-label>{{ value }}</mat-label>
         <!-- check the ngx-color-picker doc to know more about the options -->
-        <input matInput class="color-picker-input color" [formControlName]="'color'" [placeholder]="colorVal"
-            [(colorPicker)]="colorVal" [style.background]="colorVal" [style.color]="colorVal"
-            (colorPickerChange)="_handleInput(colorVal)" [cpOutputFormat]="'hex'" [cpFallbackColor]="'#ff0000'"
+        <input matInput class="color-picker-input color" [formControlName]="'color'" [placeholder]="value"
+            [(colorPicker)]="value" [style.background]="value" [style.color]="value"
+            (colorPickerChange)="_handleInput()" [cpOutputFormat]="'hex'" [cpFallbackColor]="'#ff0000'"
             [cpDisabled]="readonly" [errorStateMatcher]="matcher" readonly />
         <mat-error *ngIf="colorForm.controls.color.hasError('required')">
             Color is <strong>required</strong>
--- a/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
@@ -5,7 +5,7 @@
         <input matInput class="color-picker-input color" [formControlName]="'color'" [placeholder]="colorLabel"
             [(colorPicker)]="colorVal" [style.background]="colorVal" [style.color]="colorVal"
             (colorPickerChange)="_handleInput(colorVal)" [cpOutputFormat]="'hex'" [cpFallbackColor]="'#ff0000'"
-            [cpDisabled]="readonly" readonly />
+            [cpDisabled]="readonly" [errorStateMatcher]="matcher" readonly />
         <mat-error *ngIf="colorForm.controls.color.hasError('required')">
             Color is <strong>required</strong>
         </mat-error>
--- a/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
@@ -1,7 +1,7 @@
 <div [formGroup]="colorForm" class="color-picker-container">
     <mat-form-field appearance="outline" [floatLabel]="'always'">
         <mat-label>{{ value }}</mat-label>
-        <!-- check the ngx-color-picker doc to know more about the options -->
+        <!-- check the ngx-color-picker doc to know more about the options - https://www.npmjs.com/package/ngx-color-picker -->
         <input matInput class="color-picker-input color" [formControlName]="'color'" [placeholder]="value"
             [(colorPicker)]="value" [style.background]="value" [style.color]="value"
             (colorPickerChange)="_handleInput()" [cpOutputFormat]="'hex'" [cpFallbackColor]="'#ff0000'"
--- a/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
@@ -1,7 +1,8 @@
-<mat-form-field appearance="outline" [floatLabel]="'always'">
-    <!-- <mat-label>{{ colorLabel }}</mat-label> -->
-    <input matInput [formControlName]="'color'" [placeholder]="colorLabel" [(colorPicker)]="value.color"
-        (colorPickerChange)="_handleInput()" [cpOutputFormat]="'hex'" />
-</mat-form-field>
-
-<!-- [style.background]="colorInput" [style.color]="colorInput" -->
+<div [formGroup]="colorForm" class="color-picker-container">
+    <mat-form-field appearance="outline" [floatLabel]="'always'">
+        <mat-label>{{ colorLabel }}</mat-label>
+        <input matInput [formControlName]="'color'" [placeholder]="colorLabel" [(colorPicker)]="colorVal"
+            [style.background]="colorVal" [style.color]="colorVal" (colorPickerChange)="_handleInput(colorVal)"
+            [cpOutputFormat]="'hex'" [cpDisabled]="readonly" [cpDisableInput]="true" />
+    </mat-form-field>
+</div>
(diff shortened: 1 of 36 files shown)


projects/knora-ui/src/lib/viewer/values/color-value/color-value.component.ts
```diff
@@ -0,0 +1,115 @@
...
+    // add the submitted comment to updatedIntValue only if user has added a comment
+    if (this.commentFormControl.value !== null && this.commentFormControl.value !== '') {
+      updatedColorValue.valueHasComment = this.commentFormControl.value;
+    }
+
+    return updatedColorValue;
+  }
+
+  // update dynamically the background-color and the input label according to the picked color
+  onColorChanged(updatedValue: any) {
```
flavens: @tobiasschweizer I have set this method, onColorChanged(), to update the label and the background-color of the input before hitting the Save button (like boolean-value component but here it is the actual value old/new state, not hard-code text true/false). However the reset does not work as expected and the override method used in boolean-value component is not transposable because I cannot catch the "old" value... Any suggestion? thanks

tobiasschweizer: You should not directly manipulate/update the `displayValue`. When the update has been successfully performed, the new value fetched from Knora will be written back.

tobiasschweizer: If you change the `displayValue`, the reset does not work properly.

projects/knora-ui/src/lib/viewer/values/color-value/color-value.component.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (updatedValue) {
+      // background color of the input
+      this.displayValue.color = updatedValue;
+      // label
+      this.displayValue.strval = updatedValue;
+      // input inner text (not visible)
+      this.form.get('colorValue').setValue(updatedValue);
```
tobiasschweizer: also I do not understand why you have to set the new value in the form. You have a binding: `[formControlName]="'colorValue'"`. 

I do not understand why there is another binding: `[(colorPicker)]="displayValue.color"`

tobiasschweizer: The idea is that your input component gets its initial value from `displayValue`. Then the user can change the value. If the user saves the value, an update gets performed and the new value is written back to `displayValue`. If the user hits cancel, then the input component should be reset according to `displayValue`.

In your input component, you should only have a binding to the form. Then the input component should update the `FormControl` as the user manipulates the value. If that is not possible with a third party component, it has to be wrapped in a custom form field implementing the `ControlValueAccessor `.

tobiasschweizer:
Thanks, I will review this PR today!

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,215 @@
...
+import { Component, DoCheck, ElementRef, HostBinding, Input, OnDestroy, Optional, Self, OnInit } from '@angular/core';
+import { ControlValueAccessor, FormBuilder, FormControl, FormGroup, FormGroupDirective, NgControl, NgForm, Validators } from '@angular/forms';
+import { ErrorStateMatcher, MatFormFieldControl, mixinErrorState, CanUpdateErrorState, CanUpdateErrorStateCtor } from '@angular/material';
+import { Subject } from 'rxjs';
+import { CustomRegex } from '../../custom-regex';
+
+/**
+ * Represents a DateTime consisting of a date and a time.
+ */
+export class ColorPicker {
```
tobiasschweizer: Do we actually need an object `ColorPicker`? Couldn't this just be string? This would simplify the handling in the parent component.

projects/knora-ui/src/lib/viewer/values/color-value/color-value.component.ts
```diff
@@ -0,0 +1,124 @@
...
+  form: FormGroup;
+  valueChangesSubscription: Subscription;
+  customValidators = [];
+  matcher = new ColorErrorStateMatcher();
+
+  constructor(@Inject(FormBuilder) private fb: FormBuilder) {
+    super();
+  }
+
+  standardValidatorFunc: (val: any, comment: string, commentCtrl: FormControl) => ValidatorFn
```
tobiasschweizer: I think if you make the value a string (instead of an instance of `ColorPicker`), then you do not have to override the base class's `standardValidatorFunc`. Instead, you can just provide `customValidators = [Validators.pattern(CustomRegex.COLOR_REGEX)];`

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,215 @@
...
+  colorVal: string;
+  colorLabel: string;
+  colorForm: FormGroup;
+  colorValueFormControl: FormControl;
+  stateChanges = new Subject<void>();
+  focused = false;
+  errorState = false;
+  controlType = 'kui-color-picker';
+  matcher = new ColorPickerErrorStateMatcher();
+  colorValidator = [Validators.pattern(CustomRegex.COLOR_REGEX)];
```
tobiasschweizer: I think this is not needed here, but in the parent.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
tobiasschweizer: If it was only a string, then it would work like other components for primitive values.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  @Input() readonly = false;
+
+  @Input() errorStateMatcher: ErrorStateMatcher;
+
+  @HostBinding() id = `kui-color-picker-${ColorPickerComponent.nextId++}`;
+
+  @HostBinding('attr.aria-describedby') describedBy = '';
+
+  colorVal: string;
```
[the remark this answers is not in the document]
flavens: I kept only one variable.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
flavens: I removed the class and now work with a string.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
flavens: When I move the custom validator to the parent, my new value is always invalid. It works fine when the child component is dealing with the custom validator. But now with all your required changes, the whole validation does not work, I get weird behaviours and I do not always understand where they come from (base comp., parent comp. or child comp.)...

tobiasschweizer: Could you commit what you have done so far so I can check?

Have you removed your implementation of `standardValidatorFunc`?

tobiasschweizer: > When I move the custom validator to the parent, my new value is always invalid. It works fine when the child component is dealing with the custom validator. But now with all your required changes, the whole validation does not work, I get weird behaviours and I do not always understand where they come from (base comp., parent comp. or child comp.)...

I tried that yesterday and had some issues because the value was an object. The value now being a string, this component should work like the int component.

flavens: what do you mean by "The value now being a string"? do you mean it is working like the int component with one component and not parent/child component (with ControlValueAccessor and MatFormFieldControl)?

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
```diff
@@ -121,17 +109,17 @@ export class ColorPickerComponent extends _MatInputMixinBase implements OnInit,
...
+      return colorValue;
     }
     return null;
   }
 
-  set value(colorValue: ColorPicker | null) {
+  set value(colorValue: string | null) {
     if (colorValue !== null) {
-      this.colorForm.setValue({ color: colorValue.color });
+      this.colorForm.setValue({ color: colorValue });
```
tobiasschweizer: Just make the value a string, not an object: `this.colorForm.setValue(colorValue);`

tobiasschweizer: otherwise the getter returns an object and this is why you get a 

> Fatal error in JsonConvert. Failed to map the JavaScript instance of class "UpdateColorValue" to JSON because of a type error.

	Class property: 
		color

	Class property value: 
		[object Object]

	Expected type: 
		string

	Runtime type: 
		object

	JSON property: 
		http://api.knora.org/ontology/knora-api/v2#colorValueAsColor

	Reason: Given object does not match the expected primitive type.

tobiasschweizer: in case you need an object because of the color picker, then change the getter so it returns only the string value of the object.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
```diff
@@ -121,17 +109,17 @@ export class ColorPickerComponent extends _MatInputMixinBase implements OnInit,
   }
 
   @Input()
-  get value(): ColorPicker | null {
+  get value(): string | null {
     const colorValue = this.colorForm.value;
     if (colorValue !== null) {
-      return new ColorPicker(colorValue.color);
+      return colorValue;
```
tobiasschweizer: so this should be `colorValue.color`

tobiasschweizer:
@flavens Does representing color as a string in the sub component solve your issue?

projects/knora-ui/src/lib/viewer/values/color-value/color-value.component.spec.ts
```diff
@@ -0,0 +1,405 @@
...
+      commentInputDebugElement = valueComponentDe.query(By.css('input.comment'));
+      commentInputNativeElement = commentInputDebugElement.nativeElement;
+    });
+
+    it('should create a value', () => {
+
+      expect(testHostComponent.colorValueComponent.colorPickerComponent.value).toEqual(null);
+
+      // simulate user input
+      const newColor = '#m5m5m5';
```
tobiasschweizer: this is not valid

tobiasschweizer: character "m" is not allowed.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.html
```diff
@@ -0,0 +1,13 @@
+<div [formGroup]="colorForm" class="color-picker-container">
+    <mat-form-field appearance="outline" [floatLabel]="'always'">
+        <mat-label>{{ value }}</mat-label>
+        <!-- check the ngx-color-picker doc to know more about the options -->
```
tobiasschweizer: Could you provide a link to the docs?

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+import { By } from '@angular/platform-browser';
+
+/**
+ * Test host component to simulate parent component.
+ */
+@Component({
+  template: `
+    <div [formGroup]="form">
+      <mat-form-field>
+        <kui-color-picker #colorInput [formControlName]="'colorValue'" [readonly]="mode === 'read'"></kui-color-picker>
```
tobiasschweizer: `mode` does not exist on `TestHostComponent `

tobiasschweizer: I think this should be `[readonly]="readonly"`

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    expect(testHostComponent.form.controls.colorValue.value).toEqual('');
+  });
+
+  it('should be readonly when the readonly input is set to true', () => {
+    testHostComponent.readonly = true;
+
+    testHostFixture.detectChanges();
+
+    expect(colorInputNativeElement.readOnly).toBe(true);
```
tobiasschweizer: I think this is always readonly. I think instead you have to check if it is disabled: `[cpDisabled]="readonly"`

tobiasschweizer:
I noticed the following when running `npm install`:

> npm WARN ngx-color-picker@9.0.0 requires a peer of @angular/common@^9.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN ngx-color-picker@9.0.0 requires a peer of @angular/core@^9.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN ngx-color-picker@9.0.0 requires a peer of @angular/forms@^9.0.0 but none is installed. You must install peer dependencies yourself.

Is this version made for Angular 9?

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
flavens: I wanted to get rid of the object but the signature of setValue() expect it:
Screenshot 2020-03-11 at 14 57 23

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
flavens: it has already been updated

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
tobiasschweizer: it0s fine if you set it as an object, as long as the setter expects a string.

projects/knora-ui/src/lib/viewer/values/color-value/color-value.component.spec.ts
flavens: of course! just under my nose... too tired... thx

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
```diff
@@ -0,0 +1,190 @@
...
+  @Input() readonly = false;
+
+  @Input() errorStateMatcher: ErrorStateMatcher;
+
+  @HostBinding() id = `kui-color-picker-${ColorPickerComponent.nextId++}`;
+
+  @HostBinding('attr.aria-describedby') describedBy = '';
+
+  colorForm: FormGroup;
+  colorValueFormControl: FormControl;
```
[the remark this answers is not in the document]
flavens: has been removed

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.ts
tobiasschweizer: I think it's fine as it is now.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.spec.ts
flavens: I have to check both because the input itself should be in readonly mode all the time and the color picker opening should be disabled only in read mode.

projects/knora-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.spec.ts
```diff
@@ -98,7 +99,7 @@ describe('ColorPickerComponent', () => {
     expect(testHostComponent.form.controls.colorValue.value).toEqual('');
   });
 
-  fit('should be disabled when the readonly input is set to true', () => {
```
tobiasschweizer: Oh yes, that was the script searching for `fit` :-)

flavens: but it has been fixed in the last commit...

tobiasschweizer: I know :-) I was just happy that the script actually worked ;-)

flavens:
@tobiasschweizer should I merge it now or should we wait?!?

tobiasschweizer:
I think it's good to go. What should we wait for?

tobiasschweizer:
and do not forget to hit the checkbox in #4 :-)

PR #36, opened 2020-02-26
Time Value Component

### LINKED ISSUE #16: Viewer Value Component: TimeValueComponent
Implement `BaseValueComponent` for a `TimeValue`.

mdelez:
closes #16


mdelez:
The tests aren't working very well since my local tests rely on my local time zone and the CI operates on a different time zone so the values I check for are always wrong on one of the tests (either local tests or the CI test)

mdelez:
I think using [this](https://github.com/danielmoncada/date-time-picker) library would be ideal since it combines a date picker and a time picker. Unfortunately it requires Angular 9. There is an old version for Angular 8 but it is no longer supported.

mdelez:
Screen Shot 2020-03-04 at 12 50 17

Running the code on localhost works with no issue and returns the correct timestamp. The offset for us from UTC is +2 hours so an offset value of -120 is correct (value is in minutes and a positive offset returns a negative number). 

Screen Shot 2020-03-04 at 12 51 00

Running the unit test fails because for some reason the offset is -34. I have no idea where this seemingly random offset is coming from. From my understanding of time (which is a bit unclear now), except for a couple of timezones, namely some Australian timezones, every offset only deals with hours, not minutes or seconds. This adds to my confusion because in the tests I am setting the time value to "22:00" and it is returning a timestamp of "1776-07-04T21:25:52.000Z" when I should be expecting "1776-07-04T20:00:00.000Z" to account for the offset. Somehow the offset is also affecting the minutes and seconds.

projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.scss
```diff
@@ -0,0 +1,3 @@
+::ng-deep .parent-value-component .mat-form-field {
```
flavens: for your information: https://angular.io/guide/component-styles#deprecated-deep--and-ng-deep

mdelez: 😣technology evolves too fast. I'll update my code. Thanks!

mdelez: > ::ng-deep, /deep/ and >>> deprecation
> The ::ng-deep pseudo-class selector also has a couple of aliases: >>> and /deep/, and all three are soon to be removed.
> 
> The main reason for that is that this mechanism for piercing the style isolation sandbox around a component can potentially encourage bad styling practices.
> 
> The situation is still evolving, but right now, ::ng-deep can be used if needed for certain use cases.

https://blog.angular-university.io/angular-host-context/

From what I can find online, Angular has marked the ::ng-deep selector as deprecated but has not yet provided a new solution. They recommend to use ::ng-deep only if absolutely necessary so I shall use it sparingly.

tobiasschweizer:
@mdelez Thanks, I will look at this PR today.

package.json
```diff
@@ -24,6 +24,9 @@
     "@angular/platform-browser-dynamic": "~8.2.14",
     "@angular/router": "~8.2.14",
     "@knora/api": "file:.yalc/@knora/api",
+    "@types/lodash": "^4.14.149",
```
tobiasschweizer: is this needed because of `_.isEqual`? If so, would there be other ways to make a comparison without an additional dependency?

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,198 @@
...
+  @ViewChild('timeInput', {static: false}) timeInputComponent: TimeInputComponent;
+  @ViewChild('autosize', {static: false}) autosize: CdkTextareaAutosize;
+
+  @Input() displayValue?: ReadTimeValue;
+
+  valueFormControl: FormControl;
+  commentFormControl: FormControl;
+  date: string;
+  time: string;
+  initKnoraDate: KnoraDate; // used to display the initial value which will be an instance of KnoraDate
```
tobiasschweizer: Couldn't these be variables defined in a method's local scope? As far as I see writing and reading happens in the same place for `initKnoraDate`, `convertedControlDate`, and `updateDate`, respectively.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,198 @@
...
+export class TimeValueComponent extends BaseValueComponent implements OnInit, OnChanges, OnDestroy {
+
+  @ViewChild('timeInput', {static: false}) timeInputComponent: TimeInputComponent;
+  @ViewChild('autosize', {static: false}) autosize: CdkTextareaAutosize;
+
+  @Input() displayValue?: ReadTimeValue;
+
+  valueFormControl: FormControl;
+  commentFormControl: FormControl;
+  date: string;
```
tobiasschweizer: What are `date` and `time` used for? Could they be made local variables? 

I think updates should flow from the `displayValue` to the `FormControl`. Validation and getting the updated / new value should access solely the `FormControl`. With class members involved, I am not sure if state management works correctly.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    // when the first value is displayed on page load, it will be an instance of KnoraDate
+    if(this.valueFormControl.value.date instanceof KnoraDate){
+      this.updateDate = new Date(this.valueFormControl.value.date.year,
+                                (this.valueFormControl.value.date.month - 1),
+                                this.valueFormControl.value.date.day,
+                                splitTime[0],
+                                splitTime[1]
+      );
+    } else { // when the user submits a new value, it will be an instance of GregorianCalendarDate
```
tobiasschweizer: Could you make this consistent with the member `date` of `DateTime`? I.e. `TimeInputComponent` should receive and return a `KnoraDate` (conversion of a GregorianCalendarDate to a `KnoraDate` has to be handled by the value getter). So `TimeInputComponent` would handle the complexity of conversion.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,198 @@
...
+                       (initComment === commentFormControl.value || (initComment === null && commentFormControl.value === '')));
+
+      return invalid ? {valueNotChanged: {value: control.value}} : null;
+    };
+  };
+
+  getInitValue(): DateTime | null {
+    if (this.displayValue !== undefined) {
+      
+      const datePipe = new DatePipe('en-US');
```
tobiasschweizer: I remember that we had strange issues with timezones when using `DatePipe`. Didn't you have some issues too?

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,198 @@
...
+      if(control.value.date !== null && !(control.value.date instanceof KnoraDate)){
+        this.convertedControlDate = new KnoraDate("Gregorian",
+                                                  "AD",
+                                                  control.value.date.calendarStart.year,
+                                                  control.value.date.calendarStart.month,
+                                                  control.value.date.calendarStart.day);
+      }
+
+      const invalid = (control.value !== null &&
+                       (_.isEqual(initValue.date, this.convertedControlDate) || initValue.date === control.value.date) &&
```
tobiasschweizer: `initValue.date === control.value.date`: I think checking for equality makes only sense for primitive values, not for objects. We do not care if it is still the same object, we just care if its value is equal to what we initially had.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    // add the submitted comment to updatedTimeValue only if user has added a comment
+    if (this.commentFormControl.value !== null && this.commentFormControl.value !== '') {
+      updatedTimeValue.valueHasComment = this.commentFormControl.value;
+    }
+
+    return updatedTimeValue;
+  }
+
+  triggerResize() {
```
tobiasschweizer: Is this used?

package.json
mdelez: Removed in 046010e. I am only comparing strings now.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
mdelez: Removed in 046010e

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
mdelez: `date` and `time` have been removed in 046010e

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
mdelez: Fixed in 046010e. The child component now does all the work.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
mdelez: Removed in 046010e

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
mdelez: It is used for the textarea to automatically resize but as we've agreed I will move all the CSS and HTML changes to a new PR

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
mdelez: I haven't experienced any strange issues. The only thing is that it converts the time to the timezone of your machine.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    const newTimeValue = new CreateTimeValue();
+
+    // split the time entry in two to separate the hours and the minutes
+    let splitTime = this.valueFormControl.value.time.split(":");
+
+    // when the first value is displayed on page load, it will be an instance of KnoraDate
+    if(this.valueFormControl.value.date instanceof KnoraDate){
+      this.updateDate = new Date(this.valueFormControl.value.date.year,
```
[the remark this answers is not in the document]
mdelez: This code has been refactored in 046010e

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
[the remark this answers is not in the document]
mdelez: code moved in 9afaa90 and put into separate methods for easier testing

mdelez:
I've refactored a lot my code and I think it looks quite clean now. Namely I have moved all of the date conversion and logic to the child component and now the parent component is only concerned about a datetime string. I've also moved the date conversion logic into separate methods so that it is easier to test.

Some minor things still left to do:
- move html and css changes to separate PR
- work on the time validator so that it operates correctly within Safari since Safari doesn't support the "time" type within an input

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
```diff
@@ -0,0 +1,129 @@
...
+  constructor(@Inject(FormBuilder) private fb: FormBuilder, private _ngZone: NgZone) {
+    super();
+  }
+
+  standardValidatorFunc: (val: any, comment: string, commentCtrl: FormControl) => ValidatorFn
+    = (initValue: any, initComment: string, commentFormControl: FormControl): ValidatorFn => {
+    return (control: AbstractControl): { [key: string]: any } | null => {
+
+      const invalid = initValue === control.value &&
+                      control.value.split(':').pop().split(';')[0].match(CustomRegex.TIME_REGEX) == null &&
```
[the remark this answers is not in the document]
tobiasschweizer: In the best case overriding the `standardValidatorFunc` becomes unnecessary.

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.html
```diff
@@ -0,0 +1,22 @@
+<span [formGroup]="form" class="time-host-component">
+    <mat-form-field class="large-field child-value-component">
+      <kui-time-input #timeInput [formControlName]="'timeValue'" class="value" placeholder="Time value" [readonly]="mode === 'read'" [errorStateMatcher]="matcher"></kui-time-input>
+      <mat-error *ngIf="form.controls.timeValue.hasError('valueNotChanged')">
+        Please change the value
+      </mat-error>
+    </mat-form-field>
```
tobiasschweizer: Could you add some space or a line break  between the value and comment? 

in Chrome:
Screenshot 2020-03-12 at 12 03 36

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,231 @@
...
+              private fm: FocusMonitor,
+              private elRef: ElementRef<HTMLElement>,
+              @Optional() _parentForm: NgForm,
+              @Optional() _parentFormGroup: FormGroupDirective,
+              _defaultErrorStateMatcher: ErrorStateMatcher) { 
+
+    super(_defaultErrorStateMatcher, _parentForm, _parentFormGroup, ngControl);
+
+    this.dateFormControl = new FormControl(null);
+    this.dateFormControl.setValidators([Validators.required]);
```
tobiasschweizer: I think you can use the more compact form like below to set the validator.

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,231 @@
...
+                                userInput.date.calendarStart.day,
+                                splitTime[0],
+                                splitTime[1]
+    );
+
+    return updateDate.toISOString().split('.')[0]+"Z";
+  }
+
+  // converts and returns a unix timestamp string as an array consisting of a GregorianCalendarDate and a string
+  convertTimestampToDateTime(timestamp: string): { gcd: GregorianCalendarDate, time: string }[] {
```
tobiasschweizer: Could you make the return type an object with one member being time and the other the date?

What is `{ gcd: GregorianCalendarDate, time: string }[]` supposed to mean? An array of objects each having a member `gcd` and `time`?

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,231 @@
...
+  setDisabledState(isDisabled: boolean): void {
+    this.disabled = isDisabled;
+  }
+
+  _handleInput(): void {
+    this.onChange(this.value);
+  }
+
+  // return converted Date obj as a string without the milliseconds
+  userInputToTimestamp(userInput: any): string {
```
tobiasschweizer: Could you give the argument `userInput`a type annotation other than `any`?

tobiasschweizer: Probably you can make it consistent with what you return from `convertTimestampToDateTime`.

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,231 @@
...
+    let splitTime = userInput.time.split(":");
```
tobiasschweizer: When the user cancels the date, time seems to be an empty string:

```ts
{date: GregorianCalendarDate, time: ""}date: GregorianCalendarDate {calendarStart: CalendarDate, calendarEnd: CalendarDate, exactDate: true, jdnStart: 2458922, jdnEnd: 2458922, …}time: ""__proto__: Object

```

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  ngOnInit(): void {
+
+    this.form = this.fb.group({
+      time: '2019-07-01T12:00:00Z'
+    });
+
+  }
+}
+
+describe('TimeInputComponent', () => {
```
tobiasschweizer: Since you made separate functions for time conversion `userInputToTimestamp` and `convertTimestampToDateTime`, could add some tests for these? Just test the functions themselves, without any HTML/Dome interaction.

tobiasschweizer:
@mdelez If you need my help, just give me a call.

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: Yeah, returning an object would probably be better. Currently it's returning an array of objects because `gcd` and `time` are different types

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.ts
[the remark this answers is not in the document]
mdelez: fixed error and removed overridden validator in bfec41b

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,231 @@
...
+              @Optional() _parentForm: NgForm,
+              @Optional() _parentFormGroup: FormGroupDirective,
+              _defaultErrorStateMatcher: ErrorStateMatcher) { 
+
+    super(_defaultErrorStateMatcher, _parentForm, _parentFormGroup, ngControl);
+
+    this.dateFormControl = new FormControl(null);
+    this.dateFormControl.setValidators([Validators.required]);
+    
+    this.timeFormControl = new FormControl({value: null, Validators: [Validators.required, this.timeValidator]});
```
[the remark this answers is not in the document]
mdelez: fixed in bfec41b

projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.html
mdelez: Screen Shot 2020-03-12 at 13 46 49


It looks okay on my screen but I know the css needs some improvement. I'll move the css to a new PR and add a line break as the last thing before merging.

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in bfec41b

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in bfec41b

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in bfec41b

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.spec.ts
mdelez: tests added in bfec41b

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in 6392c37

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  // converts and returns a unix timestamp string as an array consisting of a GregorianCalendarDate and a string
+  convertTimestampToDateTime(timestamp: string): DateTime {
+    const calendarDate = new CalendarDate(Number(this.datePipe.transform(timestamp, "y")),
+                                          Number(this.datePipe.transform(timestamp, "M")),
+                                          Number(this.datePipe.transform(timestamp, "d")));
+
+    const date = new GregorianCalendarDate(new CalendarPeriod(calendarDate, calendarDate));
+
+    let time = this.datePipe.transform(timestamp, "HH:mm");
```
tobiasschweizer: use `const` instead of `let`. Please use single quotes for strings.

Or just use WebStorm .. ;-)

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,236 @@
...
+  @HostBinding('attr.aria-describedby') describedBy = '';
+
+  setDescribedByIds(ids: string[]) {
+    this.describedBy = ids.join(' ');
+  }
+
+  @Input()
+  get value(): string | null {
+    const userInput = new DateTime(this.form.value.date, this.form.value.time);
+    if (userInput.date !== null && userInput.time !== null && userInput.time !== '') {
```
tobiasschweizer: I think you should also check the time string against the regex here (to prevent an error in `this.userInputToTimestamp `). If invalid, `null` is returned.

tobiasschweizer: or probably you can just see if the `FormControl` for time is valid

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in e78d4cc I really tried to only use single quotes, but some slipped through 😜

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in 548005d

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,236 @@
...
+    if (this.form.valid) {
+      const userInput = new DateTime(this.form.value.date, this.form.value.time);
+      return this.userInputToTimestamp(userInput);
+    }
+    return null;
+  }
+
+  set value(timestamp: string | null) {
+    if (timestamp !== null) {
+      const dateTime = this.convertTimestampToDateTime(timestamp);
```
tobiasschweizer: I think this should be wrapped in a tray-catch block. If it fails to parse the timevalue string, you can still init `{date: null, time: null}`.

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  setDescribedByIds(ids: string[]) {
+    this.describedBy = ids.join(' ');
+  }
+
+  @Input()
+  get value(): string | null {
+    if (this.form.valid) {
+      const userInput = new DateTime(this.form.value.date, this.form.value.time);
+      return this.userInputToTimestamp(userInput);
```
tobiasschweizer: Could you wrap this with a try-catch block too (just return null if there is an error)?

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  _handleInput(): void {
+    this.onChange(this.value);
+  }
+
+  // return converted Date obj as a string without the milliseconds
+  userInputToTimestamp(userInput: DateTime): string {
+    const splitTime = userInput.time.split(':');
+    const updateDate = new Date(userInput.date.toCalendarPeriod().periodStart.year,
+                                (userInput.date.toCalendarPeriod().periodStart.month - 1),
```
tobiasschweizer: Please add a comment about 0-based indexes for months

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in 19c3324

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in 19c3324

projects/knora-ui/src/lib/viewer/values/time-value/time-input/time-input.component.ts
mdelez: fixed in 19c3324

PR #37, opened 2020-03-03
List Value Component

### LINKED ISSUE #15: Viewer Value Component: ListValueComponent
Implement `BaseValueComponent` for a `ListValue`.

SepidehAlassi:
To display the value:

- [x] display list node labels as nested menus

To edit the value:

- [x] edit list value as a List node 

- [x] display the nodes of the root node in a nested drop-down menu

- [x] get the list node selected by the user and assign it to the object of list node property

- [x] edit comment value

- [x] cancel edit

test:

- [x] test display

- [x] test edit

- [x] test sublist component

closes #15

### DIFF
--- a/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
+++ b/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
@@ -14,7 +14,8 @@
     </kui-color-value>
     <kui-interval-value #displayVal *ngSwitchCase="constants.IntervalValue" [mode]="mode" [displayValue]="displayValue">
     </kui-interval-value>
-    <kui-list-value #displayVal *ngSwitchCase="constants.ListValue" [mode]="mode" [displayValue]="displayValue"></kui-list-value>
+    <kui-list-value #displayVal *ngSwitchCase="constants.ListValue" [mode]="mode" [displayValue]="displayValue"
+                    [propertyDef]="parentResource.entityInfo.properties[displayValue.property]"></kui-list-value>
   </span>
 </div>
 <div>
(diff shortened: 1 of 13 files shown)


SepidehAlassi:
@tobiasschweizer I just noticed that this PR has problems with the creation of list value. So please do not review it for now.

I vaguely remember that all list nodes of a project were supposed to share a root node, is that correct? I could not find anything in that regard in knora documentation.

I know that from the lists endpoint of [api-js-lib](https://github.com/dasch-swiss/knora-api-js-lib/blob/master/src/api/v2/list/lists-endpoint.ts) one can get a list with its root node IRI. 
Now, to create a list value, I need to first find this IRI. That would require a gravsearch for a listnode with (isRootNode == true). 

My question is, should I proceed with such a gravsearch or is there a function in api-js-lib that I miss which can already return the root node?

tobiasschweizer:
> @tobiasschweizer I just noticed that this PR has problems with the creation of list value. So please do not review it for now.
> 
> I vaguely remember that all list nodes of a project were supposed to share a root node, is that correct? I could not find anything in that regard in knora documentation.
> 
> I know that from the lists endpoint of [api-js-lib](https://github.com/dasch-swiss/knora-api-js-lib/blob/master/src/api/v2/list/lists-endpoint.ts) one can get a list with its root node IRI.
> Now, to create a list value, I need to first find this IRI. That would require a gravsearch for a listnode with (isRootNode == true).
> 
> My question is, should I proceed with such a gravsearch or is there a function in api-js-lib that I miss which can already return the root node?

I think much like in the case of link values, you should find the info in the ontology:

> http://0.0.0.0:3333/ontology/0001/anything/v2#hasListItem: ResourcePropertyDefinition
id: "http://0.0.0.0:3333/ontology/0001/anything/v2#hasListItem"
subPropertyOf: ["http://api.knora.org/ontology/knora-api/v2#hasValue"]
comment: undefined
label: "Listenelement"
guiElement: "http://api.knora.org/ontology/salsah-gui/v2#List"
subjectType: "http://0.0.0.0:3333/ontology/0001/anything/v2#Thing"
objectType: "http://api.knora.org/ontology/knora-api/v2#ListValue"
isLinkProperty: false
isLinkValueProperty: false
isEditable: true
guiAttributes: ["hlist=<http://rdfh.ch/lists/0001/treeList>"]

`guiAttributes` contain the Iri of the root node, so you can the whole list with knora-api-js-lib: v2.list.getList https://dasch-swiss.github.io/knora-api-js-lib/classes/_api_v2_list_lists_endpoint_.listsendpoint.html#getlist

tobiasschweizer:
I might wanna add a cache for whole lists (currently there is just a cache for list nodes). Actually this should be combined but it is kind of tricky ...

SepidehAlassi:
@tobiasschweizer I believe this PR is ready to be reviewed.

SepidehAlassi:
@flavens thank you for reviewing this PR Flavie.

SepidehAlassi:
@flavens I've applied your suggestions.

PR #40, opened 2020-03-09
Geoname Value Component

### LINKED ISSUE #12: Viewer Value Component: GeonameValueComponent
Implement `BaseValueComponent` for a `GeonameValue`.

mdelez:
closes #12


tobiasschweizer:
will do next week!

flavens:
@mdelez @tobiasschweizer what is the state of this PR?

tobiasschweizer:
> @mdelez @tobiasschweizer what is the state of this PR?

I think I am supposed to review this. However, I am not sure whether GeoNames will be used in the future (costs for GeoNames Premium). @subotic Will we use GeoNames or is there an alternative?

flavens:
@subotic and even if we do not have projects with GeoNames in the future, are there existing projects from Salsah that will be imported and contain GeoNames we should display properly?

mdelez:
> @subotic and even if we do not have projects with GeoNames in the future, are there existing projects from Salsah that will be imported and contain GeoNames we should display properly?

I've talked to Ivan about this and he told me there are existing projects that contains Geonames so we should support them but we would also like to switch to a free alternative (the name eludes me right  now)

tobiasschweizer:
> the name eludes me right now

Open Street Map?

mdelez:
> > the name eludes me right now
> 
> Open Street Map?

That's the one :)

mdelez:
> @mdelez I have a change to suggest and additional little things:
> 
> * There is no custom validators, therefore I can enter anything in the field (string, date, decimal, etc.). Even if it will be a non-common used field and will be changed at some point, we should anyway control entered value. You should add a validator to test the format of a GeoName ID (integer, 7 characters -> to check if I am correct). This case could be tested.
> * conflicting files must be resolved
> 
> Other little things
> 
> * clean up imports in TS files (genome + display-edit)
> * in the method `openInfo()`: you should use `const` instead of `let` for the variable `url` (because the variable is never reassigned)

Thanks for the suggestions! Should I start working on them or save it for the next sprint?

flavens:
> Thanks for the suggestions! Should I start working on them or save it for the next sprint?

It depends if you have enough to work on until the end of this sprint. Otherwise, you do it.

mdelez:
> @mdelez I have a change to suggest and additional little things:
> 
> * There is no custom validators, therefore I can enter anything in the field (string, date, decimal, etc.). Even if it will be a non-common used field and will be changed at some point, we should anyway control entered value. You should add a validator to test the format of a GeoName ID (integer, 7 characters -> to check if I am correct). This case could be tested.
> * conflicting files must be resolved
> 
> Other little things
> 
> * clean up imports in TS files (genome + display-edit)
> * in the method `openInfo()`: you should use `const` instead of `let` for the variable `url` (because the variable is never reassigned)

done in 6ee361d

flavens:
@mdelez I have a suggestion about the info button (to include in this PR or to keep in mind for a design improvement)

with Angular Material, it is possible to include prefix and/or suffix inside an input
https://v8.material.angular.io/components/input/examples

I am wondering if it would be nice to place the info button as a suffix using an icon like `open_in_new`

```html
<mat-form-field class="large-field child-value-component">
    <input matInput [formControlName]="'geonameValue'" class="value" placeholder="Geoname value" type="text"
      [readonly]="mode === 'read'">
    <mat-icon matSuffix (click)="openInfo()">open_in_new</mat-icon>
</mat-form-field>
<!-- <button (click)="openInfo()">Info</button> -->
```

mdelez:
> @mdelez I have a suggestion about the info button (to include in this PR or to keep in mind for a design improvement)
> 
> with Angular Material, it is possible to include prefix and/or suffix inside an input
> https://v8.material.angular.io/components/input/examples
> 
> I am wondering if it would be nice to place the info button as a suffix using an icon like `open_in_new`
> 
> ```
> <mat-form-field class="large-field child-value-component">
>     <input matInput [formControlName]="'geonameValue'" class="value" placeholder="Geoname value" type="text"
>       [readonly]="mode === 'read'">
>     <mat-icon matSuffix (click)="openInfo()">open_in_new</mat-icon>
> </mat-form-field>
> 
> ```

Oh cool! I glanced at the prefix/suffix topic a while ago but it didn't occur to me that I could put a click event on it! I think I'll add this as a new user story though so that I can think a bit more about how the button/icon should behave :)

PR #35, opened 2020-02-26
Date Value

tobiasschweizer:
closes #11

PR #43, opened 2020-03-16
Value Component CSS and other styling

### LINKED ISSUE #39: Value components CSS
I've made some basic scss as a starting point. (As of 09.03.20 this scss is only in PR #36)

**Important**
These scss rules assume your component uses Material Icons as buttons instead of the default text buttons.

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/ede712a17bf25e8ac09671acb6c7c0fdde02a345/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html#L12-L33

I'm still using the ::ng-deep pseudo-class selector because even though the Angular team has marked it as deprecated, they have yet to provide a new solution ([it is still in the works](https://blog.angular-university.io/angular-host-context/)).

> ::ng-deep, /deep/ and >>> deprecation
> The ::ng-deep pseudo-class selector also has a couple of aliases: >>> and /deep/, and all three are soon to be removed.
> 
> The main reason for that is that this mechanism for piercing the style isolation sandbox around a component can potentially encourage bad styling practices.
> 
> The situation is still evolving, but right now, ::ng-deep can be used if needed for certain use cases.

**Changes**
I've added scss to the following files:
`display-edit.component.scss`
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/ede712a17bf25e8ac09671acb6c7c0fdde02a345/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.scss#L1-L51

These scss rules align the value, comment, and CRUD buttons horizontally, vertically align them to the top of the container and assign colors to the CRUD buttons.

`time-value.component.scss`
**These rules are only needed if you have wrapped your component**
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/ede712a17bf25e8ac09671acb6c7c0fdde02a345/projects/knora-ui/src/lib/viewer/values/time-value/time-value.component.scss#L1-L10

These rules hide the bottom line of the parent component that is generated from the `mat-form-field`

**Examples**

**Readonly**
Screen Shot 2020-03-09 at 12 11 32

**Edit (form invalid, save button greyed out)**
Screen Shot 2020-03-09 at 12 11 40

**Edit (form valid, save button enabled)**
Screen Shot 2020-03-09 at 12 22 09

mdelez:
closes #39 

- [x] Add the property label before the value form field:

```
<div>
 {{displayValue.propertyLabel}}
  <!-- info button -->
</div>
```

- [x] Display the edit icon on the right side of the value form field

- [x] Move to the value comment below the value form field

- [x] Add an "info" icon next to the property label

- [x] Change the cancel icon to the "undo" icon

- [x] Only for custom form fields with 2 inputs: align the inputs one after another and place the comment field below

- [x] Only show the comment in READ mode when user clicks on comment icon AND only if the value has a comment

- [x] Change the floating placeholder behavior of parent component value input fields to "never" 

- [x] Fix button background color on Safari & Firefox


flavens:
@mdelez after our discussion, I think we should keep the edit/save button replacement like we have now

mdelez:
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/c9380faa0531fdb8d627cd78ca5140fba78b6d21/projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-string/text-value-as-string.component.html#L5

@tobiasschweizer on a scale of 1-10 with 10 being the highest, how bad is this? For context, I am trying to show the comment for the value in READ mode only if there is a comment. Of course, when you are in UPDATE mode, the comment field must be shown regardless of its current value. Also, I'm not sure why I need to convert the mode to a string but it complains about mismatched types if I don't.

tobiasschweizer:
> https://github.com/dasch-swiss/knora-ui-ng-lib/blob/c9380faa0531fdb8d627cd78ca5140fba78b6d21/projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-string/text-value-as-string.component.html#L5
> 
> @tobiasschweizer on a scale of 1-10 with 10 being the highest, how bad is this? For context, I am trying to show the comment for the value in READ mode only if there is a comment. Of course, when you are in UPDATE mode, the comment field must be shown regardless of its current value. Also, I'm not sure why I need to convert the mode to a string but it complains about mismatched types if I don't.

I think you shouldn't complex logic in the template. Could you move `mode.toString() !== 'read' || (form.controls.comment.value !== '' && form.controls.comment.value !== null)` to the class and assign the resulting value to a class member? Then you could make a method that reevaluates the expression. You would have to see how to trigger this method.

mdelez:
> I think you shouldn't complex logic in the template. Could you move `mode.toString() !== 'read' || (form.controls.comment.value !== '' && form.controls.comment.value !== null)` to the class and assign the resulting value to a class member? Then you could make a method that reevaluates the expression. You would have to see how to trigger this method.

Maybe in the base class since every component will need it?

mdelez:
@tobiasschweizer I added a method and a variable to the base value class 

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/8cabf05c64bc512aec6515d515a2d5c4a1b24789/projects/knora-ui/src/lib/viewer/values/base-value.component.ts#L120-L122

which is then used in the template of each value component
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/8cabf05c64bc512aec6515d515a2d5c4a1b24789/projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-string/text-value-as-string.component.html#L5-L18

the method is called near the end of the resetFormControl method of the base class
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/8cabf05c64bc512aec6515d515a2d5c4a1b24789/projects/knora-ui/src/lib/viewer/values/base-value.component.ts#L100-L104

thoughts?

mdelez:
562a3e6 🤦🏻‍♂️

mdelez:
**Preface: This is only the first implementation of this. I'm sure the code can be cleaned up and simplified in one way or another.**

Based on feedback from Flavie and André, I've changed how the comments are displayed. By default the comment is always hidden in READ mode, no matter if it has a value or not. There is now a material icon to toggle the visibility of the comment in READ mode if a comment exists. When the user is in UPDATE or CREATE mode, the comment field will be displayed.

**comment hidden in READ mode**
Screen Shot 2020-03-20 at 13 22 40

**comment shown in READ mode**
Screen Shot 2020-03-20 at 13 22 50

**comment shown in UPDATE mode & comment toggle button hidden**
Screen Shot 2020-03-20 at 13 24 27

### CHANGES

**display-edit**

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/562a3e6165429ec45221857be242fb340ec44074/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.html#L11-L17

I've added a button in the template which is only shown when the user is in READ mode. The click event toggles the comment visibility.

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/562a3e6165429ec45221857be242fb340ec44074/projects/knora-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts#L102-L108

I've added two new methods in the typescript:

- **toggleComment()** calls the toggleCommentVisibility method of the value component base class which checks if the comment should be hidden or shown.

- **checkCommentToggleVisibility()** simply checks if the user is in READ mode and if the comment has a value that is not an empty string which then updates the `shouldShowCommentToggle` class variable that is used in the template to show or hide the comment toggle button. It is called in `saveEditValue` and `cancelEditValue`.

**base-value**

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/562a3e6165429ec45221857be242fb340ec44074/projects/knora-ui/src/lib/viewer/values/base-value.component.ts#L115-L128

I've added two new methods:

- **updateCommentVisibility()** updates the `shouldShowComment` class variable which is then used in the template to show or hide the comment

- **toggleCommentVisibility()** simply flips the current value of `shouldShowComment`

flavens:
@mdelez thanks for the work! After looking around, I suggest to:
- downsize the icon a bit (18px maybe) and change the blue color to black
- change the color of the label to grey (like we currently have in knora-app)
- it is possible that one property has several values with a comment, so the comment icon button cannot be placed on the property label row. It should be before or after the value form field.

mdelez:
> @mdelez thanks for the work! After looking around, I suggest to:
> 
> * downsize the icon a bit (18px maybe) and change the blue color to black
> * change the color of the label to grey (like we currently have in knora-app)
> * it is possible that one property has several values with a comment, so the comment icon button cannot be placed on the property label row. It should be before or after the value form field row.

@flavens suggestions have been added in dd73cc2 😁 For the icons, were you talking about only one icon in particular or all of the icons?

flavens:
> For the icons, were you talking about only one icon in particular or all of the icons?

I was talking about all of them. The comment icon could be first, at the left of the edit icon button.

mdelez:
> > For the icons, were you talking about only one icon in particular or all of the icons?
> 
> I was talking about all of them. The comment icon could be first, at the left of the edit icon button.

I think having the comment icon first will look a little weird if only some of the values have a comment because then the first icon will switch between the edit icon and the comment icon. If the user doesn't have write permission then the edit icon will be hidden and the comment icon will be where the edit icon usually is (first position) because of the grid layout

mdelez:
for some odd reason when I run all the unit tests on my local machine, one of the tests for the LinkValue fails but if I run that one test separately it works. Seems to work for Github CI though.

Screen Shot 2020-03-27 at 11 15 26

Screen Shot 2020-03-27 at 11 15 52

tobiasschweizer:
> for some odd reason when I run all the unit tests on my local machine, one of the tests for the LinkValue fails but if I run that one test separately it works. Seems to work for Github CI though.
> 
> Screen Shot 2020-03-27 at 11 15 26
> 
> Screen Shot 2020-03-27 at 11 15 52

#52 fixes this

flavens:
It is an empty it() test in jdndatepicker directive spec.

tobiasschweizer:
> It is an empty it() test in jdndatepicker directive spec.

will be fixed on #35

mdelez:
> > It is an empty it() test in jdndatepicker directive spec.
> 
> will be fixed on #35

this was my bad :)

SepidehAlassi:
> for some odd reason when I run all the unit tests on my local machine, one of the tests for the LinkValue fails but if I run that one test separately it works. Seems to work for Github CI though.

@mdelez exactly for this odd reason, I had not noticed that a test was failing. All the tests passed once I checked. Anyhow, I have fixed this bug in #53, can you please review it.

@tobiasschweizer #52 contained refactoring irrelevant to the bug, and I do not agree with some of the added/removed tests. I believe if a PR is opened to fix a bug, it should only contain the fix. If you wish to also refactor code or add tests, please do it in a separate PR. (that means now, please make your #52 PR to be just about refactoring)

flavens:
@tobiasschweizer would you like to take a quite look at the changed files? once this PR is merged, we will have several conflicts to fix in the other opened PRs

tobiasschweizer:
I will have a quick look tomorrow! Thanks for the notification.

mdelez:
> Could you walk me through your PR?
> I think a Google Hangout call would do it. Thanks!

Sure thing! After the daily meeting?

tobiasschweizer:
> > Could you walk me through your PR?
> > I think a Google Hangout call would do it. Thanks!
> 
> Sure thing! After the daily meeting?

OK!

PR #50, opened 2020-03-23
Text Value as Html Component

### LINKED ISSUE #6: Viewer Value Component: TextValueAsHTMLComponent
Implement BaseValueComponent for `TextValue` for a `TextValueAsHtml`.

There is only read mode.

mdelez:
closes #6

### DIFF
--- a/projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.html
@@ -1,7 +1,5 @@
+<p [innerHTML]="htmlSnippet"></p>
 <span [formGroup]="form">
-    <mat-form-field class="large-field">
-        <input matInput [formControlName]="'textValue'" class="value" placeholder="Text value" type="text" [readonly]="mode === 'read'">
-    </mat-form-field>
     <mat-form-field class="large-field">
         <input matInput [formControlName]="'comment'" class="comment" placeholder="Comment" type="text" [readonly]="mode === 'read'">
     </mat-form-field>
--- a/projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.html
@@ -1,0 +1,8 @@
+<span [formGroup]="form">
+    <mat-form-field class="large-field">
+        <input matInput [formControlName]="'textValue'" class="value" placeholder="Text value" type="text" [readonly]="mode === 'read'">
+    </mat-form-field>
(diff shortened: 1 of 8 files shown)


projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.ts
```diff
@@ -0,0 +1,66 @@
...
+      textValue: this.valueFormControl,
+      comment: this.commentFormControl
+    });
+
+    this.resetFormControl();
+  }
+
+  getInitValue() {    
+    if (this.displayValue !== undefined) {
+      this.displayValue.html = '<p>This is a very simple HTML document with a <a href="http://rdfh.ch/c9824353ae06" class="kui-link">link</a></p>';
```
tobiasschweizer: This looks like something you should do in the spec (static value). Here, you should get the HTML from the given value.

projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.ts
```diff
@@ -0,0 +1,66 @@
...
+      this.el.nativeElement.innerHTML = this.displayValue.html;
```
tobiasschweizer: How would the comment be shown?

projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.html
```diff
@@ -0,0 +1,8 @@
+<span [formGroup]="form">
+    <mat-form-field class="large-field">
+        <input matInput [formControlName]="'textValue'" class="value" placeholder="Text value" type="text" [readonly]="mode === 'read'">
```
tobiasschweizer: Probably we don't need a form field with an input if we are just going to inject HTML into the DOM.

tobiasschweizer:
@mdelez I think the some elements of the general structure don't make sense here. Probably we could discuss this later this afternoon.

projects/knora-ui/src/lib/viewer/values/text-value/text-value-as-html/text-value-as-html.component.ts
tobiasschweizer: Or could this be injected in some section in the template? I guess we need to sanitise it.

tobiasschweizer: see https://angular.io/guide/security

tobiasschweizer: Ideally we could assign the html to the value `FormControl` and sanitize it so it's rendered properly in the template.

PR #52, opened 2020-03-27
Fix Link Value Spec

tobiasschweizer:
This PR fixes a problem that occurred in the specs due to a missing spy (`searchByLabel` was triggered, but no spy was defined for the `it`).

This PR also adds some missing assertions and additional specs for link value. Checking for the active resource's label in the template is tricky. I think it's necessary to do this check because we need to know that the binding between the `FormControl` and the input element works correctly. Probably there is a better solution since we do not want to repeat tests here that are already defined for Angular Material components.

`searchByLabel`s return type should be void. This is because getting the label from Knora is an async operation (`Observable<ReadResource>`). `searchByLabel` assigns the result to `resources` when it's there and then it hets updated in the template.

### DIFF
--- a/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
+++ b/projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
@@ -1,4 +1,4 @@
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import {async, ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
 
 import { LinkValueComponent } from './link-value.component';
 import {
@@ -78,7 +78,6 @@ describe('LinkValueComponent', () => {
   beforeEach(async(() => {
     const valuesSpyObj = {
       v2: {
-        values: jasmine.createSpyObj('values', ['updateValue', 'getValue']),
         search: jasmine.createSpyObj('search', ['doSearchByLabel']),
       }
     };
@@ -108,6 +107,7 @@ describe('LinkValueComponent', () => {
     let testHostComponent: TestHostDisplayValueComponent;
     let testHostFixture: ComponentFixture<TestHostDisplayValueComponent>;
     let valueComponentDe: DebugElement;
+    let valueInputDebugElement: DebugElement;
     let valueInputNativeElement;
     let commentInputDebugElement: DebugElement;
     let commentInputNativeElement;
@@ -123,26 +123,38 @@ describe('LinkValueComponent', () => {
       const hostCompDe = testHostFixture.debugElement;
 
       valueComponentDe = hostCompDe.query(By.directive(LinkValueComponent));
-      valueInputNativeElement = valueComponentDe.query(By.css('input')).nativeElement;
+      valueInputDebugElement = valueComponentDe.query(By.css('input.value'));
+      valueInputNativeElement = valueInputDebugElement.nativeElement;
 
       commentInputDebugElement = valueComponentDe.query(By.css('input.comment'));
       commentInputNativeElement = commentInputDebugElement.nativeElement;
 
     });
 
-    it('should display an existing value', () => {
+    it('should display an existing value', fakeAsync(() => {
 
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
-      expect(testHostComponent.inputValueComponent.displayValue.propertyLabel).toMatch('Another thing');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toEqual('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.label).toEqual('Sierra');
 
       expect(testHostComponent.inputValueComponent.form.valid).toBeTruthy();
 
       expect(testHostComponent.inputValueComponent.mode).toEqual('read');
 
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.label).toEqual('Sierra');
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.type).toEqual('http://0.0.0.0:3333/ontology/0001/anything/v2#Thing');
+      expect(testHostComponent.inputValueComponent.valueFormControl.value instanceof ReadResource).toBe(true);
+      expect(testHostComponent.inputValueComponent.valueFormControl.value.label).toEqual('Sierra');
+
+      // setValue has to be called, otherwise the native input field does not get the label via the displayWith function
+      const res = testHostComponent.inputValueComponent.valueFormControl.value;
+      testHostComponent.inputValueComponent.valueFormControl.setValue(res);
+
+      // https://github.com/angular/components/blob/29e74eb9431ba01d951ee33df554f465609b59fa/src/material/autocomplete/autocomplete.spec.ts#L2577-L2580
+      testHostFixture.detectChanges();
+      tick();
+      testHostFixture.detectChanges();
+
+      expect(valueInputNativeElement.value).toEqual('Sierra');
       expect(valueInputNativeElement.readOnly).toEqual(true);
-    });
+    }));
 
     it('should make a link value editable', () => {
 
@@ -156,6 +168,7 @@ describe('LinkValueComponent', () => {
     });
 
     it('should search for resources by their label', () => {
+
       const valuesSpy = TestBed.get(KnoraApiConnectionToken);
       valuesSpy.v2.search.doSearchByLabel.and.callFake(
         () => {
@@ -166,8 +179,8 @@ describe('LinkValueComponent', () => {
         }
       );
 
-      testHostComponent.inputValueComponent.searchByLabel('thing');
-      testHostFixture.detectChanges();
+      // simulate user searching for label 'thing'
+      testHostComponent.inputValueComponent.valueFormControl.setValue('thing');
 
       expect(valuesSpy.v2.search.doSearchByLabel).toHaveBeenCalledWith('thing', 0, { limitToResourceClass: 'http://0.0.0.0:3333/ontology/0001/anything/v2#Thing'});
       expect(testHostComponent.inputValueComponent.resources.length).toEqual(1);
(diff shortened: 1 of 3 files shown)


projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -123,28 +123,40 @@ describe('LinkValueComponent', () => {
...
       commentInputDebugElement = valueComponentDe.query(By.css('input.comment'));
       commentInputNativeElement = commentInputDebugElement.nativeElement;
 
     });
 
-    it('should display an existing value', () => {
+    it('should display an existing value', fakeAsync(() => {
 
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
-      expect(testHostComponent.inputValueComponent.displayValue.propertyLabel).toMatch('Another thing');
```
SepidehAlassi: why have you removed this test?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -123,28 +123,40 @@ describe('LinkValueComponent', () => {
...
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toEqual('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.label).toEqual('Sierra');
 
       expect(testHostComponent.inputValueComponent.form.valid).toBeTruthy();
 
       expect(testHostComponent.inputValueComponent.mode).toEqual('read');
 
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.label).toEqual('Sierra');
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.type).toEqual('http://0.0.0.0:3333/ontology/0001/anything/v2#Thing');
```
SepidehAlassi: why have you removed this test?

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -123,28 +123,40 @@ describe('LinkValueComponent', () => {
...
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toMatch('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
-      expect(testHostComponent.inputValueComponent.displayValue.propertyLabel).toMatch('Another thing');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResourceIri).toEqual('http://rdfh.ch/0001/0C-0L1kORryKzJAJxxRyRQ');
+      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.label).toEqual('Sierra');
 
       expect(testHostComponent.inputValueComponent.form.valid).toBeTruthy();
 
       expect(testHostComponent.inputValueComponent.mode).toEqual('read');
 
-      expect(testHostComponent.inputValueComponent.displayValue.linkedResource.label).toEqual('Sierra');
```
SepidehAlassi: Is there any important point in moving this test from here? The one you have added above is identical.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -166,14 +201,74 @@ describe('LinkValueComponent', () => {
         }
       );
 
-      testHostComponent.inputValueComponent.searchByLabel('thing');
-      testHostFixture.detectChanges();
+      // simulate user searching for label 'thing'
+      testHostComponent.inputValueComponent.valueFormControl.setValue('thing');
```
SepidehAlassi: the point of this test was to check only the functionality of the searchByLabel directly, not through reading input value and passing it to the function. Furthermore, assigning a string 'thing' as you have done here is invalid.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -166,14 +201,74 @@ describe('LinkValueComponent', () => {
...
 
       expect(valuesSpy.v2.search.doSearchByLabel).toHaveBeenCalledWith('thing', 0, { limitToResourceClass: 'http://0.0.0.0:3333/ontology/0001/anything/v2#Thing'});
       expect(testHostComponent.inputValueComponent.resources.length).toEqual(1);
       expect(testHostComponent.inputValueComponent.resources[0].id).toEqual('http://rdfh.ch/0001/IwMDbs0KQsaxSRUTl2cAIQ');
     });
 
+    it('should not return an invalid update value (string)', () => {
```
SepidehAlassi: we have added all the validation functions to only allow input to be an instance of ReadResource, neither in update mode nor in the create mode, the input can be a string. Hence, the update function can never ever be called, we have guaranteed this with validators and there are tests for validators. This test, in my opinion, is completely unnecessary.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -194,16 +289,78 @@ describe('LinkValueComponent', () => {
 
     });
 
-    it('should return a selected resource', () => {
```
SepidehAlassi: why have you removed this test for the selected resource? It was there to check that the selected option is always an instance of ReadResuorce.

SepidehAlassi:
I removed the bug flag, as this pull request is not resolving a bug!

tobiasschweizer:
> I removed the bug flag, as this pull request is not resolving a bug!

Steps to reproduce:
- `npm run ng t --project knora-ui` (on `master`)
- check console output in Chrome:

> zone-evergreen.js:172 Uncaught TypeError: Cannot read property 'subscribe' of undefined
    at LinkValueComponent.searchByLabel (link-value.component.ts:63)
    at link-value.component.ts:105
    at SafeSubscriber.schedulerFn [as _next] (core.js:35379)
    at SafeSubscriber.__tryOrUnsub (Subscriber.js:185)
    at SafeSubscriber.next (Subscriber.js:124)
    at Subscriber._next (Subscriber.js:72)
    at Subscriber.next (Subscriber.js:49)
    at EventEmitter.next (Subject.js:39)
    at EventEmitter.emit (core.js:35298)
    at FormControl.updateValueAndValidity (forms.js:4057)

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
tobiasschweizer: I did not remove it, I just adapted some assertions. Here, the relevant things are: Iri and label

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
tobiasschweizer: I adapted the assertion so the template gets checked to. This is what we do in the other specs too. We try to keep the specs similar for all value types.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
tobiasschweizer: No, it's not in valid. Because this triggers `searchByLabel`.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
tobiasschweizer: this test makes sure that `getUpdatedValue` returns false if the form is invalid.

tobiasschweizer:
@SepidehAlassi Maybe it's easier to discuss this in a Google Hangout next week.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -260,21 +424,33 @@ describe('LinkValueComponent', () => {
...
+
       expect(testHostComponent.inputValueComponent.valueFormControl.value instanceof ReadResource).toBeTruthy();
+
       const newValue = testHostComponent.inputValueComponent.getNewValue();
       expect(newValue instanceof CreateLinkValue).toBeTruthy();
       expect((newValue as CreateLinkValue).linkedResourceIri).toEqual('http://rdfh.ch/0001/IwMDbs0KQsaxSRUTl2cAIQ');
     });
 
     it('should only create a new value if input is a resource', () => {
+      const valuesSpy = TestBed.get(KnoraApiConnectionToken);
```
SepidehAlassi: With all the changes you have applied to the other tests it is hard to see that here is where you have actually fixed the problem. Can you please dedicate one fix PR to exactly the fix instead of also applying multiple changes that are not relevant to the bug/problem at hand?

tobiasschweizer: please see https://github.com/dasch-swiss/knora-ui-ng-lib/pull/52#issuecomment-604895476

mdelez:
Is the issue I'm facing on my [PR](https://github.com/dasch-swiss/knora-ui-ng-lib/pull/43/#issuecomment-604921370) related to this?

tobiasschweizer:
> Is the issue I'm facing on my [PR](https://github.com/dasch-swiss/knora-ui-ng-lib/pull/43/#issuecomment-604921370) related to this?

I think so

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
SepidehAlassi: This kind of test (to check getUpdateValue returns false if the form is invalid) should be done for all components, not just for the link value component.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
SepidehAlassi: The label and IRI of the `linkedResource` were already checked. The test for `propertyLabel` was to check the label of the displayed value itself not the label of its `linkedResource`.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
SepidehAlassi: I had triggered the `searchByLabel` directly, what you do is triggering it while making the form invalid.

projects/knora-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
SepidehAlassi: What I mean is why you have removed the test for the `type`? there is no other test here that checks the `linkedResource.type`.

PR #56, opened 2020-04-01
Simulate productive circumstances locally

tobiasschweizer:
This PR implements https://dasch.myjetbrains.com/youtrack/issue/DSP-104

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -26,5 +26,6 @@ jobs:
         TZ: Europe/Zurich
     - run: ./find-ignored-tests.sh
     - run: npm run build-lib
+    - run: npm run build-app
       env:
         CI: true
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 To simulate circumstances of production, the application should be built with optimization and served locally 
 (not in dev mode, but from a local web server).
 
-- install `nginx` on your system, e.g. `brew install nginx` for mac OS. 
-- create a configuration for the test application.
+- install `nginx` on your system, e.g. `brew install nginx` for mac OS. See the docs for more information: <https://linux.die.net/man/8/nginx>.
+- create a configuration file for the test application.
   The example defines a configuration file `/usr/local/etc/nginx/servers/knorauiapp.conf` for macOS.
   Substitute `$abs_path_to_lib` for the actual absolute path on your system pointing to the project root.
   Substitute `$knora-ui-ng-lib_folder_name` for the folder name of the app build in `dist`.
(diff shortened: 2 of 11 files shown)


tobiasschweizer:
@flavens It's still in dev mode:

> Angular is running in the development mode. Call enableProdMode() to enable the production mode.

I figured that [production mode](https://angular.io/guide/deployment#enable-runtime-production-mode) is activated here:

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/51138c5395c2921011db56e9f0edaa9e601d8914/src/main.ts#L7-L9

It works when I set `production: true` in the dev env.

However, I tried to make a production environment so I could build it with `--prod`. But then I got some strange build errors. Could you give me a hand here?

tobiasschweizer:
@flavens Could you try out what I described in the README? 

Also I think we should really carefully check our imports. Importing from `index.ts` can cause very weird bugs. It's the IDE's fault, it should know better.

flavens:
@tobiasschweizer I will try it tomorrow!

flavens:
@tobiasschweizer I face some difficulties to follow your README.
- I modify your original script `knorauiapp.conf` on line 4 to 
`root /.../knora-ui-ng-lib/dist/knora-ui;` 
because I do not have any folder in my knora-ui-ng-lib dist named `knora-ui-ng-lib`
- when I run `nginx` from my repo '$ code/knora-ui-ng-lib', I get the following error:
`nginx: [emerg] open() "/usr/local/etc/nginx/logs/knorauiapp.local.access.log" failed (2: No such file or directory)`
I do not know how this file knorauiapp.local.access.log is supposed to be created and used...

Could you guide me a bit more?

tobiasschweizer:
> @tobiasschweizer I face some difficulties to follow your README.
> 
> * I modify your original script `knorauiapp.conf` on line 4 to
>   `root /.../knora-ui-ng-lib/dist/knora-ui;`
>   because I do not have any folder in my knora-ui-ng-lib dist named `knora-ui-ng-lib`

You can adapt the path to whatever you have on your local system. Should I make it clearer that this is a placeholder to the `dist` folder of the lib?


> * when I run `nginx` from my repo '$ code/knora-ui-ng-lib', I get the following error:
>   `nginx: [emerg] open() "/usr/local/etc/nginx/logs/knorauiapp.local.access.log" failed (2: No such file or directory)`
>   I do not know how this file knorauiapp.local.access.log is supposed to be created and used...
> 
> Could you guide me a bit more?

You just have to create an empty file, I will add that step.

tobiasschweizer:
I added the clarifications in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/56/commits/d4e769b9b1d797aeefcb69788ebe0ef664703c7a

flavens:
Thanks for the clarification!
- I would have the step-by-step process in a list with one step by bullet point and no 'first' or 'then'. I find it simpler or more efficient to follow..
- It is silly but you also could precise:

> to create the configuration file`/usr/local/etc/nginx/servers/knorauiapp.conf` and to copy the following lines and adapt the paths: ... nginx script...

- By the way, the path was correct in my case, I just renamed the subfolder of dist/. So you could write something like:
`root /$abs_path_to_lib/dist/<knora-ui-ng-lib_folder_name>;`

I went through all the steps but I cannot build the app due to an excess of budget:

Screenshot 2020-04-02 at 15 38 27

I get an 500 error in Chrome.

PR #59, opened 2020-04-15
Add missing config files

flavens:
See [YouTrack DSP-164](https://dasch.myjetbrains.com/youtrack/issue/DSP-164)


README.md
```diff
@@ -1,6 +1,142 @@
...
+* * *
+
+### knora-ui/viewer
+
+The viewer module contains object components to show the resource class representations from DSP-API, the gui-elements for the property values and different kind of view frameworks.
+[read more...](https://dasch-swiss.github.io/knora-ui/modules/viewer)
+
+* * *
+
+### knora-ui/action
```
kilchenmann, line 37: Looks good to me, but I'm not sure with the titles. We will have only one module at the end with four parts core, search, viewer and action

flavens: We can use an other term, no problem! What about "library folders"?

kilchenmann: I don't know. It's not the term "modules" that bothers me. Probably we will keep several modules in the lib? This is a general question and should be well defined or clarified in the main readme of the repo. And btw the modules are not published yet.

PR #60, opened 2020-04-15
Auto-focus on value input

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-122


mdelez:
will not be implemented

PR #63, opened 2020-04-21
Upgrade to Angular 9

flavens:
See https://dasch.myjetbrains.com/youtrack/issue/DSP-180


flavens:
@tobiasschweizer I run into this error while updating angular/material:

Screenshot 2020-04-21 at 16 19 16

what to do?

tobiasschweizer:
@flavens You have to set node version 12 in github-ci config file

tobiasschweizer:
@flavens You have to set node version 12 in github-ci config file

package.json
```diff
@@ -26,7 +26,7 @@
     "@angular/router": "~9.1.2",
     "@knora/api": "file:.yalc/@knora/api",
     "jdnconvertiblecalendar": "^0.0.5",
-    "jdnconvertiblecalendardateadapter": "^0.0.12",
+    "jdnconvertiblecalendardateadapter": "^0.0.13",
```
tobiasschweizer, line 29: @flavens I hope it works :-)

tobiasschweizer: Did it work without version conflicts this time?

tobiasschweizer:
@flavens I have updated the mock factory branch, so the rxjs version problem should be fixed. Please make sure to create package-lock.json from scratch.

package.json
flavens: yes

tobiasschweizer: cool, so it seems we know what we are doing ;-)

flavens:
@tobiasschweizer I still have an error in display-edit due to incompatible parameters in mergeMap():

Screenshot 2020-04-22 at 11 39 09

any idea?

projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.spec.ts
```diff
@@ -17,7 +18,7 @@ import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 })
 class TestHostDisplayValueComponent implements OnInit {
 
-  @ViewChild('booleanVal', { static: false }) booleanValueComponent: BooleanValueComponent;
```
tobiasschweizer, line 20: @flavens This defaults to false anyway so `{ static: false }` is not needed, right?

projects/knora-ui/src/lib/viewer/values/base-value.component.ts
```diff
@@ -1,8 +1,9 @@
-import {Input} from '@angular/core';
+import { Input, Directive } from '@angular/core';
 import {CreateValue, ReadValue, UpdateValue} from '@knora/api';
 import {AbstractControl, FormControl, FormGroup, ValidatorFn, Validators} from '@angular/forms';
 import {Subscription} from 'rxjs';
 
+@Directive()
```
tobiasschweizer, line 6: @flavens Why is this needed? It's just a class that component class extends.

projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.spec.ts
flavens: Yes, I will check the documentation and update them

tobiasschweizer: ok, then I think removing it is fine.

tobiasschweizer: I checked the docs and the sentence " Defaults to false." is new in Angular 9. So I guess it was needed before Angular 9.

projects/knora-ui/src/lib/viewer/values/base-value.component.ts
flavens: @tobiasschweizer it has been added during the migration
check https://angular.io/guide/migration-undecorated-classes

projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.spec.ts
flavens: In Angular 8, the recommendation during the upgrade from 7 to 8 was to add { static: false } for ViewChild in anticipation of Angular 9. It was only a temporary fix.

projects/knora-ui/src/lib/viewer/values/base-value.component.ts
tobiasschweizer: Thanks for this link, I was unaware of this:

> The presence of the @Directive decorator causes Angular to generate extra code for the affected class. If that decorator includes no properties (metadata), the directive won't be matched to elements or instantiated directly, but other classes that extend the directive class will inherit this generated code. You can think of this as an "abstract" directive.

So I think adding the decorator for an abstract class (cannot be instantiated) is correct.

projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.spec.ts
tobiasschweizer: Ok, got it. Thanks for the explanation!

flavens:
@tobiasschweizer the following tests are still failing where I do not find the solution:

list-value.component.spec.ts
Screenshot 2020-04-22 at 15 56 50

calendar-header.component.spec.ts
Screenshot 2020-04-22 at 15 55 45

tobiasschweizer:
> @tobiasschweizer the following tests are still failing where I do not find the solution:
> 
> list-value.component.spec.ts
> Screenshot 2020-04-22 at 15 56 50
> 
> calendar-header.component.spec.ts
> Screenshot 2020-04-22 at 15 55 45

ok, I will check it today

tobiasschweizer:
https://github.com/dasch-swiss/knora-ui-ng-lib/pull/63/commits/079944d46b5004b2ff9f7dd48502003cd6c9afe2 fixes the cal-header test

tobiasschweizer:
https://github.com/dasch-swiss/knora-ui-ng-lib/pull/63/commits/4ce02f4200141dcc3cd160d028ecc6e4c18ee330 fixes wrong signatures in `viewer/values/list-value/list-value.component.spec.ts`. I wonder why these tests worked after all. I think the assertions are insufficient, otherwise they should have failed. I think this needs a basic refactoring.

tobiasschweizer:
@flavens I talked to @mdelez and Mike will make a new branch based off his playground branch in which we merge the Angular 9 branch. So we can test everything in our browsers (we do not trust our unit tests 100% ;-)).

projects/knora-ui/src/lib/viewer/values/base-value.component.ts
tobiasschweizer: hmm, but it's a component, not a directive ...

tobiasschweizer: related question: https://github.com/angular/angular/issues/35366

tobiasschweizer: @Component inherits from @Directive(), so what we have seems right.

https://angular.io/api/core/Component

A component is essentially a directive with a template

tobiasschweizer:
@mdelez Could you check if the router you set up for the playground in Angular 8 is up to date with Angular 9?

tobiasschweizer:
I think the relevant information should be mentioned here: https://update.angular.io/

tobiasschweizer:
@mdelez I have adapted the specs from the playground in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/63/commits/3ce0c542868d505873eee33fe55cd467c03f165e so they use Angular 9 conformant code.

tobiasschweizer:
@mdelez Please check https://angular.io/guide/deprecations#loadchildren-string-syntax

mdelez:
> @mdelez Please check https://angular.io/guide/deprecations#loadchildren-string-syntax

This deprecation shouldn't have any effect on what we're currently doing as I'm doing pretty simple routing and doesn't use loadChildren :)

PR #64, opened 2020-04-23
Add Exports

tobiasschweizer:
https://dasch.myjetbrains.com/youtrack/issue/DSP-189


tobiasschweizer:
I found this post suggesting to include `index.ts` from `public-api.ts`: https://stackoverflow.com/questions/54264259/is-it-possible-for-angular-lib-to-use-nested-public-api-ts

I think for a folder `index` is implied but I got IDE warnings earlier when I just referred to the folder name without `folder/index`.

tobiasschweizer:
@mdelez I did not export the wrapped components, I think they are just an implementation detail. They are not export from the viewer module either.

tobiasschweizer:
> import { TextValueAsStringComponent } from 'knora-ui/knora-ui'; should not be a valid import path

I think 'knora-ui/knora-ui' refers to `knora-ui.d.ts` which is created during the build of the lib.

tobiasschweizer:
@mdelez This looks similar:

https://stackoverflow.com/questions/55420062/auto-importing-from-a-local-angular-library-includes-public-api-and-breaks-app

tobiasschweizer:
@mdelez Maybe there is no way to prevent someone (or an IDE) from using internal paths, but it is definitely not recommended, see https://stackoverflow.com/questions/59134983/is-import-observable-from-rxjs-internal-observable-acceptable

tobiasschweizer:
@mdelez I think we should look at the IDE settings regarding import paths, but this does seem related to this PR.

tobiasschweizer:
I think we should only have an `index.ts` on the module level. Otherwise, the IDE always tried to use `index.ts` when adding internal imports inside the lib:

`import {TextValueAsStringComponent} from './values/text-value/text-value-as-string';`

but it should be

`import {TextValueAsStringComponent} from './values/text-value/text-value-as-string/text-value-as-string.component';`

If the exports from `index.ts` are used internally, this leads to circular deps and very odd errors because the identity of the internal and external path cannot be established, and then the classes are treated as if they were different although they are not.

I am going to refactor this PR so that there is only a few `index.ts`.

mdelez:
> @mdelez Maybe there is no way to prevent someone (or an IDE) from using internal paths, but it is definitely not recommended, see https://stackoverflow.com/questions/59134983/is-import-observable-from-rxjs-internal-observable-acceptable

Yeah this seems really tricky to enforce. We might just have to make it part of the reviewing process to quickly look over newly added imports.

tobiasschweizer:
@mdelez Ok, now I have one `index.ts` per module. That should be less confusing.

I do not export classes that are considered implementation details such as wrapped components or the abstract base class.

tobiasschweizer:
I think we should consistently use relative paths although absolute paths are an option too: https://medium.com/beqode/absolute-vs-relative-import-paths-nodejs-1e4efa65a7bb

PR #71, opened 2020-05-05
cdkTextareaAutosize code cleanup

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-131

### DIFF
--- a/projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.html
+++ b/projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.html
@@ -10,7 +10,6 @@
         <mat-form-field class="large-field value-component-comment">
             <textarea matInput
                     cdkTextareaAutosize
-                    #autosize="cdkTextareaAutosize"
                     cdkAutosizeMinRows="1"
                     cdkAutosizeMaxRows="6"
                     [formControlName]="'comment'"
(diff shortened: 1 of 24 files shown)


projects/knora-ui/src/lib/viewer/values/list-value/list-value.component.ts
```diff
@@ -6,22 +6,20 @@ import {
   UpdateListValue,
   ListNodeV2,
   KnoraApiConnection,
-  ApiResponseError, ResourcePropertyDefinition
```
tobiasschweizer, line 9: you can safely remove `NgZone` from the imports since it is unused now

projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.html
```diff
@@ -11,7 +11,6 @@
             <textarea
                     matInput
                     cdkTextareaAutosize
```
tobiasschweizer, line 13: `cdkTextareaAutosize `, `cdkAutosizeMinRows`, and `cdkAutosizeMaxRows` are still needed in the template?

projects/knora-ui/src/lib/viewer/values/list-value/list-value.component.ts
mdelez: removed in a18c72d

projects/knora-ui/src/lib/viewer/values/boolean-value/boolean-value.component.html
mdelez: They're still needed in the template so that the user can see up to six lines before it will start to scroll. Otherwise the user will only have one line with a scrollbar which is not very user friendly.

tobiasschweizer: ok, got it!

PR #75, opened 2020-05-07
Error handling of modified property values 

SepidehAlassi:
- [x] mat-err message should only be visible if the user enters an invalid value. If modify is canceled, in a new try to modify values fields should be valid. 
- [x] mat-err message should be clear
- [x] every property value should have error handling for rejected value types (different error messages should be displayed for different errors returned by knora)
- [x] Simple input fields display the error message after the form is touched. Fix this! 
- [x] List Value: no validation error when the user enters the same node. Correct this!
- [x] Boolean Value: field is initially invalid if the user tries to edit
- [x] Decimal Value: Should have a custom regex to avoid entering multiple dots, and scientific notations with exponents "e" of degree higher than 20.  Explanation: Knora accepts scientific notations with 2.3en that is equal to 2.3 * 10^n if n<=20. The decimal is converted multiplying the value with 10^n. The regex pattern should prevent the user from giving exponents with n>20. 

closes [DSP-125](https://dasch.myjetbrains.com/youtrack/issue/DSP-125)


SepidehAlassi:
@mdelez I realized that the problem you mentioned for textValueAsString in [DSP-124](https://dasch.myjetbrains.com/youtrack/issue/DSP-124) is also true for interval value, link value, and time value. 
I think the previous commit should fix that. However,  I have not yet figured out the following: only for textValueAsString property, you should click somewhere else and back on the field to see the error message!

projects/dsp-ui/src/lib/viewer/values/list-value/list-value.component.ts
```diff
@@ -43,7 +43,7 @@ export class ListValueComponent extends BaseValueComponent implements OnInit, On
 
    getInitValue(): string | null {
     if (this.displayValue !== undefined) {
-      return this.displayValue.listNodeLabel;
+      return this.displayValue.listNode;
```
tobiasschweizer, line 46: I think the node should be represented as an object like in the case of other complex values (interval, time, date) implementing `ControlValueAccessor` (see https://dasch.myjetbrains.com/youtrack/issue/DSP-208).

Otherwise the component has to maintain the state of two things: label and Iri of the node. If `ControlValueAccessor` is implemented, it will be handled solely by `FormControl`. I am also thinking about using generics to make that clearer (https://dasch.myjetbrains.com/youtrack/issue/DSP-236). 

This is a good article about `ControlValueAccessor `: https://indepth.dev/never-again-be-confused-when-implementing-controlvalueaccessor-in-angular-forms/

SepidehAlassi: Refactoring working components is not the purpose of this PR.

tobiasschweizer: > Refactoring working components is not the purpose of this PR.

But there is clearly a change in list-value.component.ts, so this seems out of scope too

SepidehAlassi: > But there is clearly a change in list-value.component.ts, so this seems out of scope too

The change I have applied here is for ErrorHandling that is the purpose of this PR

tobiasschweizer: `getInitValue()` was changed because the list component is not implemented properly (it has to maintain the state of two things instead of one that would be delegated to `FormControl`.). So this should be fixed. I did not say it has to happen in this PR. But then the change of `getInitValue()` shouldn't have happened here either.

SepidehAlassi: the change of getInitValue was necessary for the `standardValueComparisonFunc` function that is necessary for ErrorHandling!

SepidehAlassi:
@mdelez & @flavens I believe this PR should catch all invalid values and display proper error messages. Please let me know if you find a case that is not handled correctly.

projects/dsp-ui/src/lib/viewer/values/time-value/time-value.component.html
```diff
@@ -7,6 +7,13 @@
   <span [formGroup]="form" class="parent-component-wrapper">
       <mat-form-field class="large-field child-value-component" floatLabel="never">
         <dsp-time-input #timeInput [formControlName]="'timeValue'" class="value" [errorStateMatcher]="matcher"></dsp-time-input>
+          <mat-error *ngIf="form.controls.timeValue.hasError('valueNotChanged') &&
+                (form.controls.timeValue.touched || form.controls.timeValue.dirty)">
+            <span class="custom-error-message">New value must be different than the current value.</span>
+          </mat-error>
+          <mat-error *ngIf="form.controls.timeValue.hasError('pattern')">
+            <span class="custom-error-message">New value must be in a valid time precision.</span>
```
mdelez, line 15: I'm having a hard time triggering this error, can you tell me what I need to do to trigger it and which browser?
Also, we could probably make this clearer by saying "A time value must follow the format HH:mm e.g. 16:30".

projects/dsp-ui/src/lib/viewer/values/value-error-state-matcher.ts
```diff
@@ -0,0 +1,9 @@
+import {FormControl, FormGroupDirective, NgForm} from '@angular/forms';
```
flavens, line 1: Formatting: space missing between { } and the imports

projects/dsp-ui/src/lib/viewer/values/time-value/time-value.component.html
SepidehAlassi: > I'm having a hard time triggering the pattern error, can you tell me what I need to do to trigger it and which browser?
> Also, we could probably make this clearer by saying "A time value must follow the format HH:mm e.g. 16:30".

to trigger it, you would need to delete the minute precision. Regarding the error message, I can make it clearer as you suggest.

projects/dsp-ui/src/lib/viewer/values/value-error-state-matcher.ts
SepidehAlassi: > Formatting: space missing between { } and the imports

I can certainly apply the rules we discussed to the new file `value-error-state-matcher`. Regarding the other files modified in this PR, I believe reformatting should be kept to a separate PR to avoid merge conflicts. what do you think?

flavens: yes exactly, we reformat only new files, and the reformatting of the others will be done later on in a specific PR

projects/dsp-ui/src/lib/viewer/values/time-value/time-value.component.html
SepidehAlassi: > I'm having a hard time triggering the pattern error, can you tell me what I need to do to trigger it and which browser?
> Also, we could probably make this clearer by saying "A time value must follow the format HH:mm e.g. 16:30".

I just noticed that this check was superfluous as the material does not allow the user to enter any invalid pattern. I made the error message for the case of the missing hour of minute precisions clearer.

mdelez: Unfortunately the check is not superfluous in Safari since what forces the format is the "type" attribute and Safari doesn't support time as a type :(

However, since we have now upgraded to Angular 9 I can use [this library](https://github.com/danielmoncada/date-time-picker) which I wanted to use when I was first making the component to supply the user with a better way to enter a time. This will of course be a refactor and in a different PR.

SepidehAlassi: So, then I trust you will add the pattern check after you update the library.

mdelez: Yeah, not sure when or if the change will happen but I will make sure to handle the pattern check.

SepidehAlassi: > Unfortunately the check is not superfluous in Safari since what forces the format is the "type" attribute and Safari doesn't support time as a type :(
> 
> However, since we have now upgraded to Angular 9 I can use [this library](https://github.com/danielmoncada/date-time-picker) which I wanted to use when I was first making the component to supply the user with a better way to enter a time. It'll also homogenize the user experience independent of the browser they are using. This will of course be a refactor and in a different PR.

Good point, I had not tried with safari. I just added the pattern check to the time-input component. In Safari, you can trigger it now with replacing `:` with `;`.

mdelez: > > Unfortunately the check is not superfluous in Safari since what forces the format is the "type" attribute and Safari doesn't support time as a type :(
> > However, since we have now upgraded to Angular 9 I can use [this library](https://github.com/danielmoncada/date-time-picker) which I wanted to use when I was first making the component to supply the user with a better way to enter a time. It'll also homogenize the user experience independent of the browser they are using. This will of course be a refactor and in a different PR.
> 
> Good point, I had not tried with safari. I just added the pattern check to the time-input component. In Safari, you can trigger it now with replacing `:` with `;`.

Looks good to me now! @flavens Do you need another look before approving?

flavens: @SepidehAlassi @mdelez no LGTM!

PR #82, opened 2020-05-13
Action module pipes migration

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-239


mdelez:
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/57bb3cd17484161432680e78f405a7cebcbaaab6/projects/dsp-ui/src/lib/action/pipes/stringify-string-literal.pipe.spec.ts#L21-L31

@tobiasschweizer can you check this spec file to make sure I understand what this pipe does and what needs to be tested?

tobiasschweizer:
> can you check this spec file to make sure I understand what this pipe does and what needs to be tested?

I have to admit that I don't know this pipe. Is this from the existing action module?

mdelez:
> > can you check this spec file to make sure I understand what this pipe does and what needs to be tested?
> 
> I have to admit that I don't know this pipe. Is this from the existing action module?

Yeah it's from the old knora-ui lib. maybe @kilchenmann  or @flavens could help?

flavens:
> Yeah it's from the old knora-ui lib. maybe @kilchenmann or @flavens could help? 

@kilchenmann did it, he is your guy on this question!

kilchenmann:
@mdelez The stringify-string-literal.pipe is documented here: https://docs.dasch.swiss/developers/knora-ui/documentation/action/stringify-string-literal/

projects/dsp-ui/src/lib/action/action.module.ts
```diff
@@ -0,0 +1,17 @@
...
+
+
+
+@NgModule({
+  declarations: [KeyPipe, ReversePipe, SortByPipe, StringifyStringLiteralPipe, TruncatePipe],
+  imports: [
+    CommonModule
+  ]
+})
+export class ActionModule { }
```
flavens, line 17: @mdelez Careful here, the name must start with Dsp, e.g. DspActionModule

mdelez: good catch! Fixed in 7341837 :)

mdelez:
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/d30bbe28e065c3269174188cf0d2eb343597b451/projects/dsp-ui/src/lib/viewer/values/boolean-value/boolean-value.component.spec.ts#L12

@tobiasschweizer I'm note sure about this import path. I moved the FormattedBooleanPipe to the action module pipes folder and now I want to access it in the boolean value component spec file.

tobiasschweizer:
> https://github.com/dasch-swiss/knora-ui-ng-lib/blob/d30bbe28e065c3269174188cf0d2eb343597b451/projects/dsp-ui/src/lib/viewer/values/boolean-value/boolean-value.component.spec.ts#L12
> 
> @tobiasschweizer I'm note sure about this import path. I moved the FormattedBooleanPipe to the action module pipes folder and now I want to access it in the boolean value component spec file.

This looks like an import path from `index.ts` which you should avoid doing. Can you try to refer to the TS file instead?

mdelez:
> > https://github.com/dasch-swiss/knora-ui-ng-lib/blob/d30bbe28e065c3269174188cf0d2eb343597b451/projects/dsp-ui/src/lib/viewer/values/boolean-value/boolean-value.component.spec.ts#L12
> > 
> > @tobiasschweizer I'm note sure about this import path. I moved the FormattedBooleanPipe to the action module pipes folder and now I want to access it in the boolean value component spec file.
> 
> This looks like an import path from `index.ts` which you should avoid doing. Can you try to refer to the TS file instead?

These are the options VS Code gives me. I don't know why or how it can come up with so many different versions.

Screen Shot 2020-05-18 at 14 52 17

tobiasschweizer:
@mdelez 

- never import from dist folder, I think you could exclude that somewhere in your IDE
- from inside the lib, do not import from `@dasch-swiss/dsp-ui`. This is how the app imports (without adding internal paths though)

the last path is correct: `../../../action/pipes/formatted-boolean-pipe`

mdelez:
> @mdelez
> 
> * never import from dist folder, I think you could exclude that somewhere in your IDE
> * from inside the lib, do not import from `@dasch-swiss/dsp-ui`. This is how the app imports (without adding internal paths though)
> 
> the last path is correct: `../../../action/pipes/formatted-boolean-pipe`

Thanks! Fixed in 6ea2da7

tobiasschweizer:
> > @mdelez
> > 
> > * never import from dist folder, I think you could exclude that somewhere in your IDE
> > * from inside the lib, do not import from `@dasch-swiss/dsp-ui`. This is how the app imports (without adding internal paths though)
> > 
> > the last path is correct: `../../../action/pipes/formatted-boolean-pipe`
> 
> Thanks! Fixed in [6ea2da7](https://github.com/dasch-swiss/knora-ui-ng-lib/commit/6ea2da74c290391129cf3b1ed949e78c6e4d088c)

I am pretty sure that you can exclude whole folders in Webstorm :-)

mdelez:
> > > @mdelez
> > > 
> > > * never import from dist folder, I think you could exclude that somewhere in your IDE
> > > * from inside the lib, do not import from `@dasch-swiss/dsp-ui`. This is how the app imports (without adding internal paths though)
> > > 
> > > the last path is correct: `../../../action/pipes/formatted-boolean-pipe`
> > 
> > 
> > Thanks! Fixed in [6ea2da7](https://github.com/dasch-swiss/knora-ui-ng-lib/commit/6ea2da74c290391129cf3b1ed949e78c6e4d088c)
> 
> I am pretty sure that you can exclude whole folders in Webstorm :-)

I know there has to be some downsides to Webstorm, it can't be just all sunshine and rainbows over there ;)

projects/dsp-ui/src/lib/action/pipes/array-transformation/sort-by.pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Pipe({
+    name: 'dspSortBy'
+})
+export class SortByPipe implements PipeTransform {
+
+    /**
+     * compares value by value and sorts by alphabetical order
+     */
+    transform(array: Array<any>, args: string): Array<any> {
```
tobiasschweizer, line 11: see my comment above about sorting

tobiasschweizer, line 11: I think an parameter should not be called  `array`.

tobiasschweizer: Is an array of objects expected and `args` is the key in an object?
If so, in case the key does not exist, it would be `undefined`, not `null`.

tobiasschweizer: better check that the param is an array, not just that it is not `undefined`

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+ *  additional characters at the end of the shortened string.
+ * The same construct with Angular slice pipe looks as follow: `{{ (str.length>24)? (str | slice:0:24)+'...':(str) }}`.
+ *
+ */
+@Pipe({
+    name: 'dspTruncate'
+})
+export class TruncatePipe implements PipeTransform {
+
+    transform(value: string, args: string[]): string {
```
tobiasschweizer, line 28: Could't the second param be numeric?

mdelez: The second param needs to be a string because you can override the default truncation string with a custom string when you give a second element in the array. Like this:
`pipe.transform('The quick brown fox jumps over the lazy dog.', ['19', '!'])`

projects/dsp-ui/src/lib/action/pipes/array-transformation/reverse.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+@Pipe({
+    name: 'dspReverse'
+})
+export class ReversePipe implements PipeTransform {
+
+    /**
+     * reverses an array
+     */
+
+    transform(value: any): any {
```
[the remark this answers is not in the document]
mdelez: I agree, sorting with a pipe seems pretty obvious at first but it seems to cause some major performance issues when scaling. Maybe this can be moved to a service so it can still be used by multiple components in the class?

projects/dsp-ui/src/lib/action/pipes/array-transformation/sort-by.pipe.ts
mdelez: can also be moved to a service

projects/dsp-ui/src/lib/action/pipes/array-transformation/key.pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+ * {{item.value.name}} and {{item.value.label}}
+ *
+ */
+@Pipe({
+    name: 'dspKey'
+})
+export class KeyPipe implements PipeTransform {
+
+    transform(value: any, args?: any): any {
+        const keys = [];
```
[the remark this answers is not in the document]
mdelez: This is difficult for me to understand. By default, all pipes are pure unless explicitly stated. What in this makes it impure? The fact that it returns a new array?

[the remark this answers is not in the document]
tobiasschweizer: >  By default, all pipes are pure unless explicitly stated. What in this makes it impure?

I think you are right and it is pure. But if you change a member in the object, you won't get an update in the template. So maybe just document that?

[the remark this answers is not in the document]
tobiasschweizer: I think you could declare it as impure and then you would get the update if just a member changes in the object.

projects/dsp-ui/src/lib/action/action.module.ts
```diff
@@ -0,0 +1,32 @@
...
+  exports: [
+    FormattedBooleanPipe,
+    KeyPipe,
+    KnoraDatePipe,
+    ReversePipe,
+    SortByPipe,
+    TruncatePipe,
+  ]
+})
+
```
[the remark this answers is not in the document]
mdelez: I don't know why but VS Code adds the blank line when I save the file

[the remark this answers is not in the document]
flavens: @mdelez in the file editorconfig, there is this option sets to true: `insert_final_newline = true`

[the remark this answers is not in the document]
mdelez: @flavens Thanks! Should I set it to false? I think this config file is version controlled right?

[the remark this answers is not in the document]
flavens: I am not sure because this is the general setup and a lot of files required a new blank line at the end...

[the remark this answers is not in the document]
mdelez: true, like the index.ts files for example

projects/dsp-ui/src/lib/action/pipes/array-transformation/sort-by.pipe.ts
flavens: @mdelez in order to avoid breaking changes, you should mark the pipe as deprecated and indicate that the service is the new way to go

projects/dsp-ui/src/lib/action/pipes/array-transformation/reverse.pipe.ts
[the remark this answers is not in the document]
mdelez: moved to a new service called `sorting.service` in 63afcde and marked the old pipe as deprecated

projects/dsp-ui/src/lib/action/pipes/array-transformation/sort-by.pipe.ts
mdelez: moved to a new service called `sorting.service` in 63afcde and marked the old pipe as deprecated

mdelez: parameter has been renamed to `value` and the check now checks for an instance of an array in 84557d0

projects/dsp-ui/src/lib/action/pipes/array-transformation/key.pipe.ts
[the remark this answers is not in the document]
mdelez: has been documented in the pipe in 63afcde

projects/dsp-ui/src/lib/action/pipes/array-transformation/key.pipe.ts
```diff
@@ -0,0 +1,26 @@
...
+ * When the value is an object with name and label, you get them with:
+ * {{item.value.name}} and {{item.value.label}}
+ *
+ */
+@Pipe({
+    name: 'dspKey'
+})
+export class KeyPipe implements PipeTransform {
+
+    transform(value: any, args?: any): any {
```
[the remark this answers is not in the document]
mdelez: @kilchenmann will check if this pipe is still needed or if the built-in pipe is enough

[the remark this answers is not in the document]
mdelez: André investigated and found that this pipe was created before there was a built-in pipe for this. This pipe has been removed in 3c8b2ae in favor of the built-in angular KeyValue pipe. The only places that use this pipe are in the viewer module list- and grid-views but this will be refactored to use the built-in pipe so I don't see this as a breaking change.

mdelez:
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/034b7f54500fb8e6a456965d7e809141ee8a2ed1/projects/dsp-ui/src/lib/action/services/sorting.service.ts#L23-L40

@kilchenmann I added an optional parameter to sort the array in reverse if desired

projects/dsp-ui/src/lib/action/pipes/array-transformation/key.pipe.ts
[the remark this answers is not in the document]
tobiasschweizer: Did you end up deleting this pipe?

projects/dsp-ui/src/lib/action/pipes/array-transformation/key.pipe.ts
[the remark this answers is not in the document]
tobiasschweizer: less code, less maintenance ;-)

projects/dsp-ui/src/lib/action/pipes/array-transformation/key.pipe.ts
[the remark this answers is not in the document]
tobiasschweizer: just read your other comment, now it's clear ;-)

projects/dsp-ui/src/lib/action/services/sorting.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    /**
+     * compares value by value and sorts by alphabetical order
+     * optionally, you can have the array returned to you in reversed order by providing the 'reversed' parameter to the method call
+     */
+    sortByAlphabetical(value: Array<any>, args: string, reversed?: boolean): Array<any> {
+        if (value instanceof Array) {
+            value.sort((a: any, b: any) => {
```
tobiasschweizer, line 25: I would be in favour of not changing the original array (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). Could you call `slice()`on the array and then work on this copy?

projects/dsp-ui/src/lib/action/services/sorting.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        if (value instanceof Array) {
+            return value.slice().reverse();
+        }
+    }
+
+    /**
+     * compares value by value and sorts by alphabetical order
+     * optionally, you can have the array returned to you in reversed order by providing the 'reversed' parameter to the method call
+     */
+    sortByAlphabetical(value: Array<any>, args: string, reversed?: boolean): Array<any> {
```
tobiasschweizer, line 23: Does array contain objects? Is it `Array<object>`?

projects/dsp-ui/src/lib/action/services/sorting.service.ts
```diff
@@ -0,0 +1,41 @@
...
+        if (value instanceof Array) {
+            value.sort((a: any, b: any) => {
+                if (args) {
+                    a[args] = (a[args] === null ? '' : a[args]);
```
tobiasschweizer, line 27: if a key is not set, it's `undefined`, not `null`. You can also use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

tobiasschweizer: `a[args] = (a[args] === null ? '' : a[args]);` modifies the original object, is that intended?

projects/dsp-ui/src/lib/action/services/sorting.service.ts
mdelez: refactored in fe5dbc8

projects/dsp-ui/src/lib/action/services/sorting.service.ts
mdelez: refactored in fe5dbc8

projects/dsp-ui/src/lib/action/services/sorting.service.ts
mdelez: refactored in fe5dbc8

tobiasschweizer:
@mdelez Thanks for this, I will have a final look now.

projects/dsp-ui/src/lib/action/services/sorting.service.ts
tobiasschweizer: looks good, I have left you some comments

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
tobiasschweizer: Could rest params be used here? https://www.typescriptlang.org/docs/handbook/functions.html#rest-parameters

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        if (!value || value.length === 0) {
+            return;
+        }
+
+        const limit = args.length > 0 ? parseInt(args[0], 10) : 20;
```
tobiasschweizer, line 34: Could you add a few comments that explain the single steps?

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+@Pipe({
+    name: 'dspTruncate'
+})
+export class TruncatePipe implements PipeTransform {
+
+    transform(value: string, args: string[]): string {
+
+        if (!value || value.length === 0) {
```
tobiasschweizer, line 30: I think you could also explicitly check like this `typeof value !== "string" || value.length === 0`

projects/dsp-ui/src/lib/action/pipes/formatting/formatted-boolean.pipe.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+        expect(convertedBoolean).toEqual('yes');
+    });
+
+    it('should return "off"', () => {
+        const myBoolean = false;
+
+        const convertedBoolean = pipe.transform(myBoolean, 'on-off');
+
+        expect(convertedBoolean).toEqual('off');
+    });
```
tobiasschweizer, line 44: are the `displayTypes` in `FormattedBooleanPipe` ever used?

mdelez: not yet, but the option has been implemented for later when the ontology editor is made

mdelez: @tobiasschweizer actually, I did use it in the boolean value template, but it's defaulting to true/false because I didn't specify a displayType

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/42c73f4895bc3f6efb202417361b92e447ea929b/projects/dsp-ui/src/lib/viewer/values/boolean-value/boolean-value.component.html#L2

tobiasschweizer:
@mdelez could you make a DSP task for the truncate pipe (see my comments)?

mdelez:
> @mdelez could you make a DSP task for the truncate pipe (see my comments)?

one step ahead of you ;) https://dasch.myjetbrains.com/youtrack/issue/DSP-276

PR #84, opened 2020-05-14
Session service migration

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-238


mdelez:
@kilchenmann is this a breaking change?

mdelez:
@tobiasschweizer I'm stuck on writing tests for this? May I ask for your assistance?

tobiasschweizer:
you most certainly may. call tomorrow at 11h?

mdelez:
https://github.com/dasch-swiss/knora-ui-ng-lib/blob/1c4a92aa4911826e58be375692d1538ec587ceeb/projects/dsp-ui/src/lib/core/session.service.spec.ts#L70

@tobiasschweizer this gives a 401 error because I am not authenticated

mdelez:
@kilchenmann some ideas for this PR:

- refactor to use cookies instead of jwt 
https://dasch.myjetbrains.com/youtrack/issue/DSP-261

- handle obtaining user information in another service so that the session service can be simplified

kilchenmann:
> refactor to use cookies instead of jwt

As we discussed last week, we have to postpone this user story

kilchenmann:
> handle obtaining user information in another service so that the session service can be simplified

This could probably be handled in dsp-js-lib in authentication service. At the moment we get only the token back and no user information. It would be great to combine it somehow...

kilchenmann:
@mdelez I'm not able to add you as a reviewer (because you created the PR). But can you have anyway a look at it? Thank.

mdelez:
> @mdelez I'm not able to add you as a reviewer (because you created the PR). But can you have anyway a look at it? Thank.

Flavie and I will review it together this afternoon!

projects/dsp-ui/src/lib/core/session.service.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        expect(service).toBeTruthy();
+    });
+
+    it('should store user information in local storage', () => {
+
+        const userSpy = TestBed.inject(DspApiConnectionToken);
+        let ls: Session;
+
+        service.setSession(undefined, 'anything.user01', 'username');
+        expect(userSpy.admin.usersEndpoint.getUser).toHaveBeenCalledTimes(1);
```
mdelez, line 81: on my side, this expect fails sometimes with this message: "Expected spy usersEndpoint.getUser to have been called once. It was called 2 times." Flavie sometimes gets the same error. Do you get this error message?

projects/dsp-ui/src/lib/core/session.service.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        expect(session.user.sysAdmin).toEqual(false);
+        expect(session.user.projectAdmin.length).toEqual(0);
+    });
+
+    it('should destroy the session', () => {
+        service.destroySession();
+        const ls: Session = JSON.parse(localStorage.getItem('session'));
+        expect(ls).toEqual(null);
+    });
+
```
mdelez, line 105: I think it would also be a good idea to test the validateSession and updateSession methods. Flavie mentioned there was an issue with testing the validateSession?

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,230 @@
...
+            if (session.id + this.MAX_SESSION_TIME <= tsNow) {
+                // the internal (dsp-ui) session has expired
+                // check if the api credentails are still valid
+
+                this.dspApiConnection.v2.auth.checkCredentials().subscribe(
+                    (response: ApiResponseData<CredentialsResponse>) => {
+                        // the knora api credentials are still valid
+
+                        // refresh the jwt in @dasch-swiss/dsp-js
+                        this.updateKnoraApiConnection(session.user.jwt);
```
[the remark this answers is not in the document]
kilchenmann: I don't remember. In this PR I was working on the test only. But, yes, it doesn't make sense.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,230 @@
...
+
+            } else {
+                // the internal (dsp-ui) session is still valid
+
+                // console.log('session is valid; return', true);
+                return true;
+            }
+        } else {
+            // no session found; update knora api connection with empty jwt
+            this.updateKnoraApiConnection();
```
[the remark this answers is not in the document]
kilchenmann: This is the KnoraAPIConfig from dsp-js-lib. We just delete the jwt there.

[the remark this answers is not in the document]
kilchenmann: https://github.com/dasch-swiss/knora-api-js-lib/blob/master/src/knora-api-config.ts

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,230 @@
...
+        );
+    }
+
+    /**
+     * Validate intern session and check knora api credentials if necessary.
+     * If a json web token exists, it doesn't mean that the knora api credentials are still valid.
+     *
+     * @returns boolean
+     */
+    validateSession(): boolean {
```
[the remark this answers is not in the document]
kilchenmann: We can rename...

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * update the session storage
+     * @param jwt
+     * @param username
+     *
+     * @returns boolean
+     */
+    updateSession(jwt: string, username: string): boolean {
```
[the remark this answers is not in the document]
kilchenmann: I don't remember. No, it doesn't make sense to return a boolean. It's probably an old setup...

projects/dsp-ui/src/lib/core/session.service.spec.ts
kilchenmann: no, never had this error message. I'm wondering...hm. But I will try again

projects/dsp-ui/src/lib/core/session.service.spec.ts
kilchenmann: I don't know how to test validateSession. How can we get a mocked jwt from fake api? I have no idea... I'm open for suggestions.

projects/dsp-ui/src/lib/core/session.service.spec.ts
kilchenmann: I tried to encapsulate the single steps. But I didn't have this error

kilchenmann: But why it runs twice sometimes? I do not understand. SOS — help needed

projects/dsp-ui/src/lib/core/session.service.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        );
+    });
+
+    it('should be created', () => {
+        expect(service).toBeTruthy();
+    });
+
+    it('should store user information in local storage', () => {
+
+        const userSpy = TestBed.inject(DspApiConnectionToken);
```
flavens, line 77: @kilchenmann if you don't use userSpy in this test, you should remove it

kilchenmann: make sense 😉

projects/dsp-ui/src/lib/core/session.service.spec.ts
flavens: @kilchenmann you try something like this?! https://stackoverflow.com/questions/41114618/trying-to-mock-angular2-token-in-test 
It is a bit old though...

mdelez: @kilchenmann @flavens I think this is good, we just need some tests that check: 

- if isSessionValid returns false when there is no session
- if isSessionValid returns true when the internal dsp-ui session is still valid
- if isSessionValid calls `dspApiConnection.v2.auth.checkCredentials()` when the dsp-ui session is expired
    - For this, we will also need to mock the auth service. I don't know how to mock 2 services in the same test since they are both used for the `DspApiConnectionToken` in the providers

kilchenmann: I will try to write something...

projects/dsp-ui/src/lib/core/session.service.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+        (valuesSpy.admin.usersEndpoint as jasmine.SpyObj<UsersEndpointAdmin>).getUser.and.callFake(
+            () => {
+                console.log('getting a fake user');
+
+                const loggedInUser = MockUsers.mockUser();
+                return of(loggedInUser);
+            }
+        );
+
+        (valuesSpy.v2.auth as jasmine.SpyObj<AuthenticationEndpointV2>).checkCredentials.and.callFake(
```
mdelez, line 75: @tobiasschweizer this mocked method is never called when trying to run the test on line 135

tobiasschweizer: ok, I will have a look at this.

tobiasschweizer: I think I found out why the mock is not called. It is because inside the session service a new `KnoraApiConnection` is instantiated:

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/5f175699b3534732946ab85f7695de4d7d29151b/projects/dsp-ui/src/lib/core/session.service.ts#L198-L206

I think instead the `KnoraApiConnection` provided in the constructor should be used. I don't understand why a new connection object has to be constructed.

tobiasschweizer:
@mdelez I think I can do the review on Thursday this week.

tobiasschweizer:
@mdelez Ok, I will look at this now.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        localStorage.removeItem('session');
+    }
+
+    /**
+     * Update the dsp-api-config and dsp-api-connection of @dasch-swiss/dsp-js
+     *
+     * @param  {string} jwt?
+     */
+    private updateDspApiConnection(jwt?: string) {
+        this.dspApiConfig.jsonWebToken = (jwt ? jwt : '');
```
tobiasschweizer, line 204: Isn't it enough to set the jwt on the `DspApiConfigToken`?

tobiasschweizer: private methods should start with an underscore

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+    sysAdmin: boolean;
+
+    // list of project shortcodes where the user is project admin
+    projectAdmin: string[];
+}
+
+/**
+ * Session with id (= login timestamp) and inforamtion about logged-in user
+ */
+export interface Session {
```
tobiasschweizer, line 36: if `Session` is only used internally, there is no need to export it.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+    KnoraApiConfig,
+    KnoraApiConnection,
+    UserResponse
+} from '@dasch-swiss/dsp-js';
+import { DspApiConfigToken, DspApiConnectionToken } from './core.module';
+
+/**
+ * Currently logged-in user information
+ */
+export interface CurrentUser {
```
tobiasschweizer, line 16: if `CurrentUser` is only used internally, there is no need to export it.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+    Constants,
+    CredentialsResponse,
+    KnoraApiConfig,
+    KnoraApiConnection,
+    UserResponse
+} from '@dasch-swiss/dsp-js';
+import { DspApiConfigToken, DspApiConnectionToken } from './core.module';
+
+/**
+ * Currently logged-in user information
```
tobiasschweizer, line 14: Information about the current user?

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * get session information from localstorage
+     */
+    getSession(): Session {
+        if (localStorage.getItem('session') !== null) {
+            return JSON.parse(localStorage.getItem('session'));
+        }
+
+        return undefined;
```
tobiasschweizer, line 67: functions return `undefined` by default, no need for this.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    constructor(
+        @Inject(DspApiConnectionToken) private dspApiConnection: KnoraApiConnection,
+        @Inject(DspApiConfigToken) private dspApiConfig: KnoraApiConfig
+    ) { }
+
+    /**
+     * get session information from localstorage
+     */
+    getSession(): Session {
```
tobiasschweizer, line 62: I would be better to give this function an explicit return type. If you cannot find user information, return `null`, e.g. `Session | null`.

projects/dsp-ui/src/lib/core/session.service.ts
tobiasschweizer: better return `null`

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+                    }
+
+                    if (response.body.user.permissions.groupsPerProject[key].indexOf(Constants.ProjectAdminGroupIRI) > -1) {
+                        projectAdmin.push(key);
+                    }
+                }
+
+                // store session information in browser's localstorage
+                // TODO: jwt will be removed, when we have a better cookie solution (DSP-261)
+                session = {
```
tobiasschweizer, line 106: Make a constructor for `CurrentUser` and `Session` and instantiate the class using `new`.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+                        name: response.body.user.username,
+                        jwt: jwt,
+                        lang: response.body.user.lang,
+                        sysAdmin: sysAdmin,
+                        projectAdmin: projectAdmin
+                    }
+                };
+
+                // update localStorage
+                localStorage.setItem('session', JSON.stringify(session));
```
tobiasschweizer, line 118: this is only executed once the answer from Knora arrives. So after calling `setSession` there is no way to know when this is done.

Could this be be done in a better way? For example, the method could return an `Observable` but there could be a `pipe` that handles writing the information to the cache.

tobiasschweizer: with RxJS, you can perform actions without subscribing: https://www.learnrxjs.io/learn-rxjs/operators/utility/do

The subscriber would be the one that calls `setSession`. Once the subscriber gets notified, the session has been set.

projects/dsp-ui/src/lib/core/session.service.ts
tobiasschweizer: Or maybe just make it an interface since you have to deserialize it from json. It would be good to add a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types). TypeScript has a structural type system anyway. So you could juts check if the object has the necessary properties after deserialisation.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        );
+    }
+
+    /**
+     * Validate intern session and check knora api credentials if necessary.
+     * If a json web token exists, it doesn't mean that the knora api credentials are still valid.
+     *
+     * @returns boolean
+     */
+    isSessionValid(): boolean {
```
tobiasschweizer, line 133: I think since this method involves a call to the Knora API, it can only return an `Observable<boolean>`. In the branches where no connection to Knora is necessary, you can return of(true) or `of(false)`.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+                this.dspApiConnection.v2.auth.checkCredentials().subscribe(
+                    (response: ApiResponseData<CredentialsResponse>) => {
+                        // the knora api credentials are still valid
+
+                        // update the session.id
+                        session.id = tsNow;
+
+                        localStorage.setItem('session', JSON.stringify(session));
+
+                        return true;
```
tobiasschweizer, line 157: I think this cannot possibly work correctly.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * update the session storage
+     * @param jwt
+     * @param username
+     *
+     * @returns boolean
+     */
+    updateSession(jwt: string, username: string) {
```
tobiasschweizer, line 184: This method is not used anywhere. Can it be removed?

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        if (session) {
+
+            this.updateDspApiConnection(session.user.jwt);
+
+            // check if the session is still valid:
+            if (session.id + this.MAX_SESSION_TIME <= tsNow) {
+                // the internal (dsp-ui) session has expired
+                // check if the api credentials are still valid
+
+                this.dspApiConnection.v2.auth.checkCredentials().subscribe(
```
tobiasschweizer, line 148: rather than subscribing, perform the necessary actions with `pipe`. The caller of this method should subscribe.

tobiasschweizer:
The methods `setSession` and `isSessionValid` should return an `Observable`.

projects/dsp-ui/src/lib/core/session.service.ts
```diff
@@ -0,0 +1,215 @@
...
+    providedIn: 'root'
+})
+export class SessionService {
+
+    /**
+     * max session time in milliseconds
+     * default value (24h): 86400000
+     *
+     */
+    readonly MAX_SESSION_TIME: number = 86400000; // 1d = 24 * 60 * 60 * 1000
```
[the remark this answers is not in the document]
kilchenmann: No, this is an internal session time. It's not the same as in Knora. This one is valid for 24 hours. Only after this time, a new authentication request is made to Knora. Session time in Knora is about 30 days?

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: No, we have to use it in the app too. It has to be exported.

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Same as above. We need the information of the currently logged in user. It's important to know, if a user is system admin or not. And in which project a user is project admin.

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: That's true.

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: > Isn't it enough to set the jwt on the DspApiConfigToken?

There is this property `jsonWebToken` in the dspApiConfig provided by dsp-js-lib. I thought it's a good idea to assign. When it's not needed and we do not have to set it, we can delete this line. But why do we have this `jsonWebToken` property in dspApiConfig?

tobiasschweizer: the jwt is set on the `KnoraApiConfig`, see https://github.com/dasch-swiss/knora-api-js-lib/blob/bc169c0403b19d81096be81d871ef3c12ff29549/src/api/endpoint.ts#L35-L47

tobiasschweizer:
I think there might be a timing problem on application startup. I am trying to find a way to make the `DspApiConfigToken` and `DspApiConnectionToken` providers dependent on `initializeApp`. See https://medium.com/@gmurop/managing-dependencies-among-app-initializers-in-angular-652be4afce6f

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Yes, dspApiConfig is of type KnoraApiConfig. So, we don't have to set it?

kilchenmann: dspApiConfig is initiated in AppInitService and can be used in the ui-lib. Described in the ui-lib README: 
https://www.npmjs.com/package/@dasch-swiss/dsp-ui#setup

tobiasschweizer: For some reason there are problems with the config token, see  https://github.com/dasch-swiss/knora-ui-ng-lib/pull/118. Whenever I try to use it ina component or a service, I get an error (dependency injection).

projects/dsp-ui/src/lib/core/session.service.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+                return store[key] || null;
+            }
+        );
+        spyOn(localStorage, 'removeItem').and.callFake(
+            (key: string): void => {
+                delete store[key];
+            }
+        );
+        spyOn(localStorage, 'setItem').and.callFake(
+            (key: string, value: string): string => {
```
tobiasschweizer, line 60: `setItem` does not return a value, see https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: I saw that there's a missing part in CoreModule. Please have a look at the old knora-ui core module:
https://github.com/dasch-swiss/knora-ui/blob/master/projects/knora/core/src/lib/core.module.ts

```typescript
    static forRoot(kuiConfig: KuiConfig): ModuleWithProviders {
        // get the app environment configuration here
        // console.log('KuiCoreModule - forRoot - config: ', config);
        return {
            ngModule: KuiCoreModule,
            providers: [
                { provide: KuiConfigToken, useValue: kuiConfig }
            ]
        };
    }
```

kilchenmann: I will fix it...

tobiasschweizer: could you try that on https://github.com/dasch-swiss/knora-ui-ng-lib/pull/118?

tobiasschweizer: You should mention the type, see https://angular.io/guide/migration-module-with-providers:

```
@NgModule({...})
export class MyModule {
  static forRoot(config: SomeConfig): ModuleWithProviders<SomeModule> {
    return {
      ngModule: SomeModule,
      providers: [
        {provide: SomeConfig, useValue: config }
      ]
    };
  }
}
```

kilchenmann: > could you try that on #118?

Yes I will

Makefile
```diff
@@ -74,6 +72,33 @@ knora-stack: ## runs the knora-stack
 	sleep 35
 	$(MAKE) -C $(CURRENT_DIR)/.tmp/knora-stack stack-logs-api-no-follow
 
+.PHONY: prepare-test-env
```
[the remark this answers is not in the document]
kilchenmann: Probably yes. I will delete it...

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: If a system admin removes himself as system admin we have to update the user data in the local storage. The same applies to project admin permissions. A user creates a project and becomes project admin himself. In this case the local storage has to be updated so that the GUI knows what it is allowed to display in the project area and what not.

kilchenmann: Will be resolved in the app [DSP-449](https://dasch.myjetbrains.com/youtrack/issue/DSP-449)

kilchenmann: Resolved in [6c9bc68](https://github.com/dasch-swiss/knora-ui-ng-lib/pull/84/commits/6c9bc68bd1e0eb715973d7ee48df8675202883c3)

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: > private methods should start with an underscore

Resolved in [6c9bc68](https://github.com/dasch-swiss/knora-ui-ng-lib/pull/84/commits/6c9bc68bd1e0eb715973d7ee48df8675202883c3)

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: As the `CurrentUser` is part of `Session` we should export `Session` only.

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: > Isn't it enough to set the jwt on the `DspApiConfigToken`?

I agree now. Resolved in 75c1377

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Resolved in cb139e0

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Resolved in d272070

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Resolved in c3eb394

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Resolved in cb139e0

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: > I saw that there's a missing part in CoreModule. Please have a look at the old knora-ui core module:
> https://github.com/dasch-swiss/knora-ui/blob/master/projects/knora/core/src/lib/core.module.ts
 
I'm sorry, but my answer was nonsense. This is a leftover from the old knora-ui setup and doesn't make sense anymore. I have tested it in #118 but have already undone it.

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Yes

projects/dsp-ui/src/lib/core/session.service.ts
kilchenmann: Part of this conversation is resolved in PR #118

tobiasschweizer:
@kilchenmann @mdelez I have merged #118 into this branch. Could we check after this afternoon's meeting if this PR is ready to go?

kilchenmann:
>  I have merged #118 into this branch. Could we check after this afternoon's meeting if this PR is ready to go?

Yes, that's fine. Thanks!

src/app/action-playground/action-playground.component.ts
```diff
@@ -105,6 +115,46 @@ export class ActionPlaygroundComponent implements OnInit {
     }
 
     sortList(key) {
-        this.list = this.sortingService.keySortByAlphabetical(this.list, key);
+        this.list = this._sortingService.keySortByAlphabetical(this.list, key);
     }
+
+    // TODO: Will be replaced by login process from action module
+    /*
+    login() {
```
tobiasschweizer, line 123: So here we will use the login / logout component instead?

projects/dsp-ui/src/lib/core/session.service.ts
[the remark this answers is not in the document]
tobiasschweizer: It really depends on the settings in Knora. I don't think you can make any assumptions about this in the client.

projects/dsp-ui/src/lib/core/session.service.ts
tobiasschweizer: Could you change the comment accordingly to `// information about the current user`?

projects/dsp-ui/src/lib/core/session.service.spec.ts
tobiasschweizer: `setItem` has a return type of `void`,  `store[key] = value` is enough.

tobiasschweizer:
@kilchenmann Mike and I will take care of the remaining tasks.

src/app/action-playground/action-playground.component.ts
kilchenmann: Yes exactly. It doesn't exist yet in the ui-lib action module. The login/logout component will use the session service

kilchenmann:
> @kilchenmann Mike and I will take care of the remaining tasks.

Thanks

tobiasschweizer:
Merging of this PR will have to wait until @subotic checked the setup.

tobiasschweizer:
@subotic Regarding the issue you mentioned I could imagine that there was a problem with services that were provided at the module level:

> We have an Angular 6 Library that is just domain services that I want others in the org to use. Currently, configured like MyModule.forRoot( environment.whateverConfig ) in app.module.ts. The issue is some of the settings in environment.whateverConfig need to be provided by a call to the server for which we use APP_INITIALIZER. The problem is forRoot() is executed before the promise in the APP_INITIALIZER provider finishes, so the library does not get the correct config.

https://github.com/angular/angular/issues/23279#issuecomment-412108311

In `dsp-ui`, all services are provided on the root level. If for some reason we need to provide some services on the module level, this could be an issue. The call of `forRoot` on the module is necessary to prevent multiple instances of the same service if the module is imported multiple times (if provided on the module level), see https://angular.io/guide/singleton-services#the-forroot-pattern. As the comment mentions, `forRoot` is executed before APP_INITIALIZER provider finishes which is a problem because the configuration which is fetched async is not available.

tobiasschweizer:
@subotic https://stackoverflow.com/questions/43193049/app-settings-the-angular-way

PR #88, opened 2020-05-25
Refactor how to get properties by type

flavens:
For more info: [DSP-142](https://dasch.myjetbrains.com/youtrack/issue/DSP-142) and [DSP-275](https://dasch.myjetbrains.com/youtrack/issue/DSP-275)


tobiasschweizer:
As long as there isn't a new RC for the JS lib, it has to be built manually. 
From the JS lib's project root:

- `npm run prepare-dev-publication` (includes mock files)
- `npm run yalc-publish`

tobiasschweizer:
@flavens Hopefully Non Breaking Nes: @knora/api@1.0.0-rc.1 is out!!!

package.json
```diff
@@ -31,7 +31,7 @@
     "@angular/platform-browser": "~9.1.2",
     "@angular/platform-browser-dynamic": "~9.1.2",
     "@angular/router": "~9.1.2",
-    "@knora/api": "1.0.0-rc.0",
+    "@knora/api": "1.0.0-rc.1",
```
tobiasschweizer, line 34: Please do not forget to adapt the lib's package.json to "1.0.0-rc.1"

flavens: @tobiasschweizer sorry but I do not understand this comment

tobiasschweizer: The lib has its own `package.json`: https://github.com/dasch-swiss/knora-ui-ng-lib/blob/wip/DSP-142-getPropList/projects/dsp-ui/package.json

Your code requires a specific version of `@knora/api` and because of this it needs to be adapted in the lib's peer deps. When using the lib in another Angular app than the test app, npm can then check if the app env provides the version the lib needs.

there is a good article about this: https://indepth.dev/create-your-standalone-angular-library-in-10-minutes/

tobiasschweizer, line 34: Could you also commit the change in `package-lock.json`? It still says 1.0.0-rc.0.

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
```diff
@@ -60,38 +55,23 @@ export class ResourceViewComponent implements OnInit, OnChanges {
       (response: ReadResource) => {
```
tobiasschweizer, line 55: the param docstring of getResource says resource instead of iri.

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
```diff
@@ -24,40 +31,34 @@ class TestPropertyParentComponent implements OnInit {
...
-          }
-
-        } else {
-          console.error('Error detected: the property with IRI =' + index + 'is not a property of the resource');
-        }
-      }
+        this.parentResource = response;
+
+        // gather resource property information
+        this.propArray = this.parentResource.entityInfo.classes[this.parentResource.type].getResourcePropertiesList().map(
```
tobiasschweizer, line 42: Do I understand correctly that you have to mock the information that would be passed in from `ResourceViewComponent`? And you do this in `ResourceViewComponent` and not in `PropertyViewComponent` because otherwise the same thing would have to be repeated for each property?

I have a similar problem in the JS lib: the mock involves some complexity. I think your approach is totally fine for now, but could you create a DSP task so we can evaluate:
- whether to centralise this in UI lib (this would also involve some thinking about whether in specs we should use productive code or not at all)
- move this logic to JS lib to so the information would be returned with a mocked resource

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
```diff
@@ -1,8 +1,8 @@
 <div class="resource-view" *ngIf="resource">
 
   <!-- dsp-property-view -->
-  <div *ngIf="propArray.length !== 0; else noProperty">
-    <dsp-property-view [parentResource]="resource" [propArray]="propArray" [systemPropArray]="systemPropArray">
+  <div *ngIf="resPropInfoVals.length !== 0 || systemPropDefs.length !== 0; else noProperty">
```
tobiasschweizer, line 4: Is this check necessary? How would `PropertyViewComponent` behave if both `resPropInfoVals` and `systemPropDefs` are empty arrays?

You don't have to worry about not initialised members because bot arrays are initialised with an empty array. I think that's a good practice.

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
```diff
@@ -1,6 +1,13 @@
...
+import {
+    ApiResponseError,
+    IHasPropertyWithPropertyDefinition,
+    MockResource,
+    ReadResource,
+    ReadValue,
+    SystemPropertyDefinition
+} from '@knora/api';
 import { PropertyInfoValues } from '../resource-view/resource-view.component';
 import { PropertyViewComponent } from './property-view.component';
```
tobiasschweizer, line 13: a comment referring to `knora-ui-ng-lib/projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.ts`:

```ts
  /**
   * Array of property object with ontology class prop def, list of properties and corresponding values
   *
   * @param (propArray)
   */
  @Input() propArray: PropertyInfoValues;

  /**
   * Array of system property object with list of system properties
   *
   * @param (propArray)
   */
  @Input() systemPropArray?: PropertyDefinition;
```

Shouldn't these be arrays `PropertyInfoValues[]` and `PropertyDefinition[]`? Also I think `systemPropArray` needn't be optional, the array could just be empty.

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
```diff
@@ -60,38 +55,23 @@ export class ResourceViewComponent implements OnInit, OnChanges {
...
 
         // sort properties by guiOrder
-        this.propArray.sort((a, b) => (a.guiDef.guiOrder > b.guiDef.guiOrder) ? 1 : -1);
+        this.resPropInfoVals.sort((a, b) => (a.guiDef.guiOrder > b.guiDef.guiOrder) ? 1 : -1);
+
+        // get system property information
+        this.systemPropDefs = this.resource.entityInfo.getPropertyDefinitionsByType(SystemPropertyDefinition);
 
       },
       (error: ApiResponseError) => {
```
tobiasschweizer, line 77: Good that you check for unsuccessful responses, I think we should apply such a pattern throughout the lib / app.

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
flavens: ok so I will remove this check

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
flavens: resolve in commit [b452f0e](https://github.com/dasch-swiss/knora-ui-ng-lib/pull/88/commits/b452f0e6304f8b8665c15d593c51a6b5398a18c1)

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
flavens: what do you mean?

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
flavens: Yes, you understood it correctly. I even asked for your opinion on this when I created the mock and it tests and you agreed. I can create a DSP issue, no problem.

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
tobiasschweizer: the documentation string for the method

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
flavens: [DSP-282](https://dasch.myjetbrains.com/youtrack/issue/DSP-282)

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
flavens: ok, the related line to your comment misled me...

flavens: resolved in 1972f45

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
tobiasschweizer: Ok, fine with me if you leave this check.

tobiasschweizer: I mean, the current way it is implemented is fine :-)

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
tobiasschweizer: >  I even asked for your opinion on this when I created the mock and it tests and you agreed.

ok :-)

PR #89, opened 2020-05-26
Refactor Truncate Pipe

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-276

### DIFF
--- a/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.spec.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.spec.ts
@@ -42,7 +42,7 @@ describe('TruncatePipe', () => {
         const fixture = TestBed.createComponent(App);
         fixture.detectChanges();
 
-        expect(fixture.nativeElement.textContent).toBe('This...');
+        expect(fixture.nativeElement.textContent).toEqual('This...');
     });
 
     it('should support the trail argument in the template', () => {
@@ -57,6 +57,6 @@ describe('TruncatePipe', () => {
         const fixture = TestBed.createComponent(App);
         fixture.detectChanges();
 
-        expect(fixture.nativeElement.textContent).toBe('This is!!');
+        expect(fixture.nativeElement.textContent).toEqual('This is!!');
     });
 });
--- a/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.spec.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.spec.ts
@@ -2,9 +2,11 @@ import { TruncatePipe } from './truncate.pipe';
 
 describe('TruncatePipe', () => {
     let pipe: TruncatePipe;
+    let snippet: string;
 
     beforeEach(() => {
         pipe = new TruncatePipe();
+        snippet = 'The quick brown fox jumps over the lazy dog.';
     });
 
     it('create an instance', () => {
@@ -12,10 +14,17 @@ describe('TruncatePipe', () => {
     });
 
     it('should truncate after 15 characters', () => {
-        expect(pipe.transform('The quick brown fox jumps over the lazy dog.', ['15'])).toEqual('The quick brown...');
+        const truncatedSnippet = pipe.transform(snippet, 15);
+        expect(truncatedSnippet).toEqual('The quick brown...');
     });
 
     it('should truncate after 19 characters and add an exclamation mark at the end', () => {
-        expect(pipe.transform('The quick brown fox jumps over the lazy dog.', ['19', '!'])).toEqual('The quick brown fox!');
+        const truncatedSnippet = pipe.transform(snippet, 19, '!');
+        expect(truncatedSnippet).toEqual('The quick brown fox!');
+    });
+
+    it('should truncate after 20 characters by default', () => {
+        const truncatedSnippet = pipe.transform(snippet);
+        expect(truncatedSnippet).toEqual('The quick brown fox ...');
     });
 });
--- a/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.spec.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.spec.ts
@@ -1,4 +1,6 @@
 import { TruncatePipe } from './truncate.pipe';
+import { Component } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
 
 describe('TruncatePipe', () => {
     let pipe: TruncatePipe;
@@ -26,5 +28,35 @@ describe('TruncatePipe', () => {
     it('should truncate after 20 characters by default', () => {
         const truncatedSnippet = pipe.transform(snippet);
         expect(truncatedSnippet).toEqual('The quick brown fox ...');
+    });
+
+    it('should support the limit argument in the template', () => {
+        @Component({
+          template: `{{ text | dspTruncate:4 }}`,
+        })
+        class App {
+          text = 'This is my string';
+        }
+
+        TestBed.configureTestingModule({declarations: [App, TruncatePipe]});
+        const fixture = TestBed.createComponent(App);
+        fixture.detectChanges();
+
+        expect(fixture.nativeElement.textContent).toBe('This...');
+    });
+
+    it('should support the trail argument in the template', () => {
+        @Component({
+          template: `{{ text | dspTruncate:7:'!!' }}`,
+        })
+        class App {
+          text = 'This is my string';
+        }
+
+        TestBed.configureTestingModule({declarations: [App, TruncatePipe]});
+        const fixture = TestBed.createComponent(App);
+        fixture.detectChanges();
+
+        expect(fixture.nativeElement.textContent).toBe('This is!!');
     });
 });
--- a/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
@@ -24,19 +24,22 @@ import { Pipe, PipeTransform } from '@angular/core';
 })
 export class TruncatePipe implements PipeTransform {
 
+    defaultLimit = 20;
+    defaultTrail = '...';
+
     transform(value: string, limit?: number, trail?: string): string {
 
         if (typeof value !== 'string' || value.length === 0) {
             return '';
         }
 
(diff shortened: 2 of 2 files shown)


mdelez:
Would it make sense to return an empty string instead of undefined?

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L29-L31

mdelez:
Could rest params be used here? https://www.typescriptlang.org/docs/handbook/functions.html#rest-parameters

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L24-L27

mdelez:
Could you add a few comments that explain the single steps?

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L33-L35

mdelez:
I think you could also explicitly check like this `typeof value !== "string" || value.length === 0`

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L29-L31

mdelez:
> Would it make sense to return an empty string instead of undefined?
> 
> https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L29-L31

@tobiasschweizer fixed in 00b6a60

mdelez:
> I think you could also explicitly check like this `typeof value !== "string" || value.length === 0`
> 
> https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L29-L31

@tobiasschweizer fixed in 00b6a60

mdelez:
> Could rest params be used here? https://www.typescriptlang.org/docs/handbook/functions.html#rest-parameters
> 
> https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L24-L27

@tobiasschweizer I don't think this is really what we want since it's not combined a lot of strings. It's just truncating and adding something at the end. I've decided to make the pipe take two arguments related to the truncation, a number for the length and a string for the text to add at the end in 00b6a60

mdelez:
> Could you add a few comments that explain the single steps?
> 
> https://github.com/dasch-swiss/knora-ui-ng-lib/blob/a2ed2f59aa87e79913069aadc9c878192dce0903/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L33-L35

@tobiasschweizer https://github.com/dasch-swiss/knora-ui-ng-lib/blob/00b6a60f36e0a7081b45d6041450ea660c4a5593/projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts#L27-L40

I refactored it to be easier to read. Doesn't look nearly as cool though.

mdelez:
@kilchenmann opinions welcome :)

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
```diff
@@ -25,14 +24,18 @@ import { Pipe, PipeTransform } from '@angular/core';
...
-        if (!value || value.length === 0) {
-            return;
+        if (typeof value !== 'string' || value.length === 0) {
+            return '';
         }
 
-        const limit = args.length > 0 ? parseInt(args[0], 10) : 20;
-        const trail = args.length > 1 ? args[1] : '...';
+        // default to 20 if no character limit is provided
+        if (!limit) { limit = 20; }
```
tobiasschweizer, line 34: I think it is not a good practice to reassign to a parameter (https://eslint.org/docs/rules/no-param-reassign).

I think you could do the following:

```ts
export class TruncatePipe implements PipeTransform {

    defaultLimit = 20;
    defaultTrail = '...';

    transform(value: string, limit?: number, trail?: string): string {

        if (typeof value !== 'string' || value.length === 0) {
            return '';
        }

        const limitSetting: number = limit !== undefined ? limit : this.defaultLimit;
        const trailSetting: string = trail !== undefined ? trail : this.defaultTrail;

        return value.length > limitSetting ? value.substring(0, limitSetting) + trailSetting : value;
    }

}
```

The ternary operators are not easy to understand but quite elegant. Maybe just add some comments?

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
```diff
@@ -25,14 +24,18 @@ import { Pipe, PipeTransform } from '@angular/core';
 })
 export class TruncatePipe implements PipeTransform {
 
-    transform(value: string, args: string[]): string {
+    transform(value: string, limit?: number, trail?: string): string {
```
tobiasschweizer, line 27: I was actually wondering why this works since the interface looks like this:

```ts
export declare interface PipeTransform {
    transform(value: any, ...args: any[]): any;
}
```

So `args` is actually a rest parameter which you can access as an array. However, I tested your implementation from the template and it works.

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
mdelez: Ha, yeah that was a rookie move on my part. I should know better 😅

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
tobiasschweizer: I have found this:

https://github.com/angular/angular/blob/a62c1c441c67616ce5eacc7e8dcfe169316d4552/packages/core/test/acceptance/pipe_spec.ts#L22-L27

tobiasschweizer: I wonder if we should test pipes in templates, see https://github.com/angular/angular/blob/a62c1c441c67616ce5eacc7e8dcfe169316d4552/packages/core/test/acceptance/pipe_spec.ts#L86-L99

mdelez: > I have found this:
> 
> https://github.com/angular/angular/blob/a62c1c441c67616ce5eacc7e8dcfe169316d4552/packages/core/test/acceptance/pipe_spec.ts#L22-L27

I believe this is what I'm already doing, right?

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
mdelez: rookie mistake has been fixed in e5491d7

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.ts
tobiasschweizer: > I believe this is what I'm already doing, right?

yes

Do you think it makes sense to test the pipe in a template setup:

```ts
 it('should support arguments in pipes', () => {
    @Component({
      template: `{{person.name | multiArgPipe:'one':person.address.city}}`,
    })
    class App {
      person = {name: 'value', address: {city: 'two'}};
    }

    TestBed.configureTestingModule({declarations: [App, MultiArgPipe]});
    const fixture = TestBed.createComponent(App);
    fixture.detectChanges();

    expect(fixture.nativeElement.textContent).toBe('value one two default');
  });
```

https://github.com/angular/angular/blob/a62c1c441c67616ce5eacc7e8dcfe169316d4552/packages/core/test/acceptance/pipe_spec.ts#L86-L99

mdelez: > > I believe this is what I'm already doing, right?
> 
> yes
> 
> Do you think it makes sense to test the pipe in a template setup:
> 
> ```ts
>  it('should support arguments in pipes', () => {
>     @Component({
>       template: `{{person.name | multiArgPipe:'one':person.address.city}}`,
>     })
>     class App {
>       person = {name: 'value', address: {city: 'two'}};
>     }
> 
>     TestBed.configureTestingModule({declarations: [App, MultiArgPipe]});
>     const fixture = TestBed.createComponent(App);
>     fixture.detectChanges();
> 
>     expect(fixture.nativeElement.textContent).toBe('value one two default');
>   });
> ```
> 
> https://github.com/angular/angular/blob/a62c1c441c67616ce5eacc7e8dcfe169316d4552/packages/core/test/acceptance/pipe_spec.ts#L86-L99

Couldn't hurt to add

mdelez: pipe via template tests added in bbd61c9

tobiasschweizer: Wow!

projects/dsp-ui/src/lib/action/pipes/string-transformation/truncate.pipe.spec.ts
```diff
@@ -1,21 +1,62 @@
...
+        })
+        class App {
+          text = 'This is my string';
+        }
+
+        TestBed.configureTestingModule({declarations: [App, TruncatePipe]});
+        const fixture = TestBed.createComponent(App);
+        fixture.detectChanges();
+
+        expect(fixture.nativeElement.textContent).toBe('This...');
```
tobiasschweizer, line 45: very small remark: I prefer `toEqual` (equivalence) since `toBe` checks for identify. 

https://stackoverflow.com/questions/22413009/jasmine-javascript-testing-tobe-vs-toequal

But it does not matter in case of primitive values.

mdelez: I agree, has been changed to `toEqual` in 365750d

PR #90, opened 2020-05-28
Int Value Comp Specs: Using MatHarness

tobiasschweizer:
relates to https://dasch.myjetbrains.com/youtrack/issue/DSP-140

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/values/int-value/int-value.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/values/int-value/int-value.component.spec.ts
@@ -82,12 +82,8 @@ describe('IntValueComponent', () => {
         let testHostComponent: TestHostDisplayValueComponent;
         let testHostFixture: ComponentFixture<TestHostDisplayValueComponent>;
         let valueComponentDe: DebugElement;
-        let valueInputDebugElement: DebugElement;
-        let valueInputNativeElement;
         let valueReadModeDebugElement: DebugElement;
         let valueReadModeNativeElement;
-        let commentInputDebugElement: DebugElement;
-        let commentInputNativeElement;
 
         let loader: HarnessLoader;
 
@@ -125,9 +121,6 @@ describe('IntValueComponent', () => {
 
             testHostFixture.detectChanges();
 
-            valueInputDebugElement = valueComponentDe.query(By.css('input.value'));
-            valueInputNativeElement = valueInputDebugElement.nativeElement;
-
             expect(testHostComponent.inputValueComponent.mode).toEqual('update');
 
             expect(testHostComponent.inputValueComponent.form.valid).toBeFalsy();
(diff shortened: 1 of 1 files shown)


tobiasschweizer:
@mdelez Because of a bug in `MatIn putHarness`, one `it` still uses the old way of getting the DOM element. See https://github.com/angular/components/issues/18790#issuecomment-635206117

tobiasschweizer:
@mdelez I have a thought regarding the mock resource:

```ts
    ngOnInit() {

        MockResource.getTestthing().subscribe(res => {
            const inputVal: ReadIntValue =
                res.getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasInteger', ReadIntValue)[0];

            this.displayInputVal = inputVal;

            this.mode = 'read';
        });

    }
```

This is async (in theory at least because in the test env it behaves as if sync): we do not really know when the result is there. I think maybe I could change this Observable to a Promise and then use `await`.

mdelez:
> @mdelez I have a thought regarding the mock resource:
> 
> ```ts
>     ngOnInit() {
> 
>         MockResource.getTestthing().subscribe(res => {
>             const inputVal: ReadIntValue =
>                 res.getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasInteger', ReadIntValue)[0];
> 
>             this.displayInputVal = inputVal;
> 
>             this.mode = 'read';
>         });
> 
>     }
> ```
> 
> This is async (in theory at least because in the test env it behaves as if sync): we do not really know when the result is there. I think maybe I could change this Observable to a Promise and then use `await`.

I'm not really sure of the advantage of changing the observable to a promise is. Aren't observables almost always preferred?

tobiasschweizer:
> I'm not really sure of the advantage of changing the observable to a promise is. Aren't observables almost always preferred?

Your are right, an Observable should not be converted to a Promise. However, with a Promise you can use the async await syntax that's easier to read than a callback.

What I am trying to figure out is if the fact that our mock data is returned as an Observable could lead to unpredictable timing issues in the test.

However, in the Angular docs I found the mention of a "synchronous Observable" (https://angular.io/guide/testing#synchronous-tests). So then the test would actually be synchronous.

tobiasschweizer:
I think I just have to make sure that the `Observable` returned by `MockResource.getTestthing()` behaves like a "synchronous Observable" .

If this is the case, everything can stay as is. No premature optimisation ...

tobiasschweizer:
maybe this helps: https://medium.com/angular-in-depth/how-to-test-observables-a00038c7faad

mdelez:
> I think I just have to make sure that the `Observable` returned by `MockResource.getTestthing()` behaves like a "synchronous Observable" .
> 
> If this is the case, everything can stay as is. No premature optimisation ...

We should at some  point also incorporate error handling and testing, which it seems isn't possible with a synchronous test:
`You're not so lucky when testing the error path. Although the service spy will return an error synchronously, the component method calls setTimeout(). The test must wait at least one full turn of the JavaScript engine before the value becomes available. The test must become asynchronous.`

tobiasschweizer:
> We should at some point also incorporate error handling and testing, which it seems isn't possible with a synchronous test:

Yes, absolutely! I try to figure this out tomorrow.

tobiasschweizer:
Interesting read: https://netbasal.com/testing-observables-in-angular-a2dbbfaf5329

tobiasschweizer:
Ok, this works:

```ts
ngOnInit() {

        const delay = 0;

        const test$ = MockResource.getTestthing().pipe(observeOn(queueScheduler, delay));

        test$.subscribe(res => {
            const inputVal: ReadIntValue =
                res.getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasInteger', ReadIntValue)[0];

            this.displayInputVal = inputVal;

            this.mode = 'read';
        });
}
```

and this does not:

```ts

const test$ = MockResource.getTestthing().pipe(observeOn(asyncScheduler, delay));

        test$.subscribe(res => {
           
            const inputVal: ReadIntValue =
                res.getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasInteger', ReadIntValue)[0];

            this.displayInputVal = inputVal;

            this.mode = 'read';
        });

```

Note the different scheduler: `queueScheduler ` -> `asyncScheduler `

mdelez:
> Ok, this works:
> 
> ```ts
> ngOnInit() {
> 
>         const delay = 0;
> 
>         const test$ = MockResource.getTestthing().pipe(observeOn(queueScheduler, delay));
> 
>         test$.subscribe(res => {
>             const inputVal: ReadIntValue =
>                 res.getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasInteger', ReadIntValue)[0];
> 
>             this.displayInputVal = inputVal;
> 
>             this.mode = 'read';
>         });
> }
> ```
> 
> and this does not:
> 
> ```ts
> const test$ = MockResource.getTestthing().pipe(observeOn(asyncScheduler, delay));
> 
>         test$.subscribe(res => {
>            
>             const inputVal: ReadIntValue =
>                 res.getValuesAs('http://0.0.0.0:3333/ontology/0001/anything/v2#hasInteger', ReadIntValue)[0];
> 
>             this.displayInputVal = inputVal;
> 
>             this.mode = 'read';
>         });
> ```
> 
> Note the different scheduler: `queueScheduler ` -> `asyncScheduler `

Nice! Maybe it would be better to make these changes in another PR as I don't think it directly relates to using the harness?

tobiasschweizer:
> Nice! Maybe it would be better to make these changes in another PR as I don't think it directly relates to using the harness?

Yes, I think I will merge this PR as is. I just wanted to find out if the current behaviour is what we need.

Because if the `MockResource.getTestthing()` does not behave synchronously as expected, then the tests could behave in a non deterministic way, leading to non reproducible fails.

PR #91, opened 2020-05-28
Migrate action module directives

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-294


mdelez:
@flavens @kilchenmann this PR is just to migrate the directives from the old action module to the new one. Refactoring (if needed) will happen in another PR :)

@kilchenmann can you also have a look at the existing-name spec file? I don't understand the need for the regex stuff:

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/600174646fe049b3aa7ca39c6fb15fbd6821b3cd/projects/dsp-ui/src/lib/action/directives/existing-names/existing-name.directive.spec.ts#L37-L39

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/600174646fe049b3aa7ca39c6fb15fbd6821b3cd/projects/dsp-ui/src/lib/action/directives/existing-names/existing-name.directive.spec.ts#L56-L63

kilchenmann:
>  I don't understand the need for the regex stuff

It's needed for a string comparison.

PR #94, opened 2020-06-03
Migrate sort button

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-320

### DIFF
--- a/projects/dsp-ui/src/lib/action/components/sort-button/sort-button.component.spec.ts
+++ b/projects/dsp-ui/src/lib/action/components/sort-button/sort-button.component.spec.ts
@@ -1,0 +1,169 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { MatIconModule } from '@angular/material/icon';
+import { MatMenuModule } from '@angular/material/menu';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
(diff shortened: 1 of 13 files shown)


mdelez:
Screen Shot 2020-06-03 at 11 00 25

In addition to moving the sort button, I also decided to make a playground for these action components :)

projects/dsp-ui/src/lib/action/components/sort-button/sort-button.component.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+        TestBed.configureTestingModule({
+            imports: [
+                MatIconModule,
+                MatMenuModule,
+                BrowserAnimationsModule
+            ],
+            declarations: [
+                SortButtonComponent,
+                TestHostComponent,
+                SortByPipe]
```
flavens, line 101: @mdelez why do you still use the sort-by pipe marked as deprecated? would it make sense to use the new sorting service instead (as explained in the sort-by pipe annotation)?

kilchenmann: same question here 😉

mdelez:
> LGTM

You guys are fast! Thanks!! @flavens @kilchenmann

PR #95, opened 2020-06-03
Migration of message component

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-327

### DIFF
--- a/projects/dsp-ui/src/lib/action/components/message/message.component.html
+++ b/projects/dsp-ui/src/lib/action/components/message/message.component.html
@@ -1,0 +1,1 @@
+<p>message works!</p>
--- a/projects/dsp-ui/src/lib/action/components/message/message.component.html
+++ b/projects/dsp-ui/src/lib/action/components/message/message.component.html
@@ -1,1 +1,46 @@
-<p>message works!</p>
+<mat-card *ngIf="!short" class="fix-width dsp-message" [ngClass]="'dsp-' + message?.type">
+
+    <mat-card-subtitle class="message-subtitle">
+        <span class="left">{{message?.type | uppercase }} {{message?.status}} | {{message?.statusMsg}}</span>
+        <!-- <mat-icon class="right">{{message?.type}}</mat-icon> -->
+    </mat-card-subtitle>
+
+    <mat-card-title class="message-title" [innerHtml]="message?.statusText"></mat-card-title>
+
+    <mat-card-content class="message-content">
+        <mat-list *ngIf="message?.route">
+            <mat-list-item>
+                <a [href]="message?.route" target="_blank">
+                    &rarr; {{message?.route}}
+                </a>
+            </mat-list-item>
+        </mat-list>
+
+        <mat-list *ngIf="showLinks">
+            <p>{{links.title}}</p>
+            <mat-list-item *ngFor="let item of links.list" class="link" (click)="goToLocation(item.route)">
+                <mat-icon mat-list-icon>{{item.icon}}</mat-icon>
+                <p mat-line>{{item.label}}</p>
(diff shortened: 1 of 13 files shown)


mdelez:
Screen Shot 2020-06-04 at 16 09 58

Not sure why it's giving this warning. It should be included in the action module.

kilchenmann:
> Screen Shot 2020-06-04 at 16 09 58
> 
> Not sure why it's giving this warning. It should be included in the action module.

Were you able to fix this?

mdelez:
> > Screen Shot 2020-06-04 at 16 09 58
> > Not sure why it's giving this warning. It should be included in the action module.
> 
> Were you able to fix this?

Yeah I fixed it in eaee147. For some reason I was using the component in another template and didn't realize. I think I accidentally copy and pasted it.

projects/dsp-ui/src/lib/action/components/message/message.component.ts
```diff
@@ -0,0 +1,208 @@
...
+     * @param short Show short message only
+     * It can be used in form to show if a post was successfull or not.
+     */
+    @Input() short = false;
+
+    /**
+     * @param medium Show medium message
+     * A message box without footnote or links.
+     */
+    @Input() medium = false;
```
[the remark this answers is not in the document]
mdelez: I think that's smart! Otherwise someone could use both parameters and I'm not sure how the system would handle that. For the names, how about: "short", "medium", "long"?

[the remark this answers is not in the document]
mdelez: added TODO comment in dc48a9e

projects/dsp-ui/src/lib/action/components/message/message.component.ts
```diff
@@ -0,0 +1,210 @@
...
+    statusMsg: any;
+
+    isLoading = true;
+
+    showLinks = false;
+
+    // disable message
+    disable = false;
+
+    /*
```
flavens, line 61: there is a * missing

projects/dsp-ui/src/lib/action/components/message/message.component.html
```diff
@@ -0,0 +1,46 @@
...
+
+<mat-card *ngIf="short && !disable" class="fix-width dsp-short-message" [ngClass]="'dsp-' + message?.type"
+          (click)="closeMessage()">
+
+    <div class="dsp-panel">
+        <span class="dsp-short-message-text">
+            {{message?.statusText}}
+        </span>
+        <span class="fill-remaining-space"></span>
+        <!--         <button mat-icon-button (click)="closeMessage()">
```
flavens, line 41: do we need to keep this commented code block??

projects/dsp-ui/src/lib/action/components/message/message.component.ts
mdelez: added in 8acfb4d :)

projects/dsp-ui/src/lib/action/components/message/message.component.html
mdelez: Most likely we don't but I wanted to keep everything as close to identical to the old code as possible. This was commented out in the old code as well. I would assume it's safe to remove?

flavens: I suppose yes
@kilchenmann can we remove this code block?

kilchenmann: It was an icon `x` to close the message. But I see that a click event on the whole card is set to close the message. So, we don't need the icon anymore. You can delete this commented code block...

PR #96, opened 2020-06-03
test (test inputs): experiment with default values for inputs

tobiasschweizer:
This PR adds a test component to the app that illustrates how to use default values for `@Input`s.


src/app/development/input-test/input-test.component.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+import { Component, OnInit, ViewChild } from '@angular/core';
+
+/**
+ * Test host component to simulate parent component.
+ */
+@Component({
+    template: `
+        <app-input-test #testComp [myFirstInput]="'dada'"></app-input-test>`
+})
+class TestHostComponent1 implements OnInit {
```
mdelez, line 13: Screen Shot 2020-06-04 at 10 52 54

Just a small linter issue, but since this is just a testing component I don't think it's very important to change

src/app/development/input-test/input-test.component.spec.ts
```diff
@@ -0,0 +1,191 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { InputTestComponent } from './input-test.component';
+import { Component, OnInit, ViewChild } from '@angular/core';
```
mdelez, line 4: Imports should be in alphabetical order 😛

tobiasschweizer: ok, will fix this :-)

src/app/development/input-test/input-test.component.spec.ts
tobiasschweizer: I think we can leave it as is.

src/app/development/input-test/input-test.component.spec.ts
tobiasschweizer: done in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/96/commits/2d8fcc115fdc796cb21faaa37e910531f87c7e65

PR #98, opened 2020-06-04
Set Up E2E Tests

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-311


tobiasschweizer:
@mdelez I think when I add more tests, I can probably add more convenience methods. So testing simple cases will require a method call.

PR #100, opened 2020-06-09
Use strictTemplates Compiler Option

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-319

~~requires #101~~ 

This PR makes the lib compile with Ivy using `"strictTemplates": true`. Compilation problems are ignored using `$any()` (https://angular.io/guide/template-syntax#the-any-type-cast-function).
However, the following issues still have to be fixed:
- ~~https://dasch.myjetbrains.com/youtrack/issue/DSP-393 (should be addressed soon)~~
- ~~https://dasch.myjetbrains.com/youtrack/issue/DSP-391 (trivial)~~
- https://dasch.myjetbrains.com/youtrack/issue/DSP-389 (not urgent)

This PR splits tests into groups:
- unit tests and productive build of lib and app (required)
- dev build of lib
- E2E tests (required)

I did not make  dev build of lib required since we do not use it in production (Ivy is not used to build the lib for production). However, this test should pass. More checks by the compiler should lead to fewer problems at runtime, that's my rationale.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,14 +9,15 @@ env:
   KNORA_GDB_LICENSE: ${{ github.workspace }}/ci/graphdb/UNIBAS_DEV_GRAPHDB_SE_latest-31-12-2020_1cores.license
   GRAPHDB_HEAP_SIZE: 4G
 
+strategy:
+  matrix:
+    node-version: [12.x]
+
 jobs:
   build-test:
     name: Build and Test
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [12.x]
 
     steps:
     - uses: actions/checkout@v2
@@ -40,9 +41,6 @@ jobs:
     name: Build lib in dev mode
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [12.x]
 
     steps:
       - uses: actions/checkout@v2
@@ -59,9 +57,6 @@ jobs:
     name: E2E Tests
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [12.x]
 
     steps:
     - uses: actions/checkout@v2
@@ -80,8 +75,8 @@ jobs:
     - run: |
         make clone-knora-stack
         make knora-stack
-        npm run webdriver-update
-        npm run e2e
+    - run: npm run webdriver-update
+    - run: npm run e2e
 
 
   publish:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,30 +37,37 @@ jobs:
           CI: true
     - run: npm run build-lib-dev
 
+
+
   e2e-test:
     name: E2E Tests
     needs: [build-test]
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x]
+
     steps:
-      - uses: actions/checkout@v2
-      - name: Checkout private ci-assets
-        uses: actions/checkout@v2
-        with:
-          repository: dasch-swiss/dsp-ci-assets
-          token: ${{ secrets.ci_assets_pat }} # `GitHub_PAT` is a secret that contains the PAT.
-          path: ci
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
-        with:
-          node-version: ${{ matrix.node-version }}
-      - run: npm install
-      - run: npm run build-lib
-      - run: |
-          sudo apt-get install expect
-          make clone-knora-stack
-          make knora-stack
-          npm run webdriver-update
-          npm run e2e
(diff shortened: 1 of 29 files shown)


tobiasschweizer:
@mdelez I am going to group tests on this PR, so tests are run individually (like in https://github.com/dasch-swiss/knora-api/blob/f984c2e0589aec81e28c45adde4ff76d9820b640/.github/workflows/main.yml#L62-L142). And not all of them would be required for merging, so we could detect problems without blocking development.

.github/workflows/main.yml
```diff
@@ -41,12 +35,57 @@ jobs:
...
           TZ: Europe/Zurich
           CI: true
+
+  build-dev:
+    name: Build lib in dev mode
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x]
```
tobiasschweizer, line 45: @subotic Do you happen to know if there could be one central version matrix? This would be easier to update.

tobiasschweizer:
@mdelez I made different groups of tests, and not all of them are required.

mdelez:
> @mdelez I made different groups of tests, and not all of them are required.

Perfect!

.github/workflows/main.yml
subotic: You could try and see if [defaults](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaults) could allow you to define a value which you then could use in all the different matrix statements.

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.html
```diff
@@ -11,21 +11,21 @@ <h4 class="label">
...
-      <dsp-decimal-value class="parent-value-component" #displayVal *ngSwitchCase="constants.DecimalValue" [mode]="mode" [displayValue]="displayValue"></dsp-decimal-value>
-      <dsp-color-value class="parent-value-component" #displayVal *ngSwitchCase="constants.ColorValue" [mode]="mode" [displayValue]="displayValue"></dsp-color-value>
-      <dsp-interval-value class="parent-value-component" #displayVal *ngSwitchCase="constants.IntervalValue" [mode]="mode" [displayValue]="displayValue"></dsp-interval-value>
-      <dsp-time-value class="parent-value-component" #displayVal *ngSwitchCase="constants.TimeValue" [mode]="mode" [displayValue]="displayValue"></dsp-time-value>
-      <dsp-geoname-value class="parent-value-component" #displayVal *ngSwitchCase="constants.GeonameValue" [mode]="mode" [displayValue]="displayValue"></dsp-geoname-value>
-      <dsp-link-value class="parent-value-component" #displayVal *ngSwitchCase="constants.LinkValue" [mode]="mode" [displayValue]="displayValue"
+      <dsp-text-value-as-string class="parent-value-component" #displayVal *ngSwitchCase="'ReadTextValueAsString'" [mode]="mode" [displayValue]="$any(displayValue)"></dsp-text-value-as-string>
```
mdelez, line 14: should I work on making a service that returns the correct value type? It shouldn't take too long and then you can use the service in this branch so we don't have to use temporary code that may get left in the master branch for a while

tobiasschweizer: I think the other two issues are more important. Could we do them first?

tobiasschweizer: In case of the `DisplayEditComponent`, we know that this will be correct at runtime. We just have to tell the compiler somehow.

.github/workflows/main.yml
tobiasschweizer: will do, thx

tobiasschweizer:
not to myself: make e2e tests required when this PR has been merged.

projects/dsp-ui/src/lib/action/components/sort-button/sort-button.component.html
```diff
@@ -2,7 +2,7 @@
     <button mat-icon-button [matMenuTriggerFor]="sortSelection">
         <mat-icon>{{icon}}</mat-icon>
     </button>
-    <mat-menu #sortSelection="matMenu" [xPosition]="$any(menuXPos)">
+    <mat-menu #sortSelection="matMenu" [xPosition]="menuXPos">
```
tobiasschweizer: @mdelez 👍

tobiasschweizer:
@kilchenmann @mdelez @flavens Could you have a look at this PR? I'd like to merge it before working on migrating the advanced search, so that the new code is compiled in strict template mode.

tobiasschweizer:
@mdelez I think it is ok to use `$any()` in `DisplayEditCompopnent`'s template for the moment because we know the correct type will be passed at runtime.

PR #105, opened 2020-06-10
Move the search-panel and the fulltext-search components

flavens:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-392


projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+    });
+
+    it('should get projects on init', () => {
+        const projSpy = TestBed.inject(DspApiConnectionToken);
+
+        expect(testHostComponent.fulltextSearch.projects).toBeDefined();
+        expect(testHostComponent.fulltextSearch.projects.length).toEqual(8);
+        expect(testHostComponent.fulltextSearch.projectfilter).toEqual(true);
+
+        // TO FIX: expect(projSpy.admin.projectsEndpoint.getProjects).toHaveBeenCalledTimes(1);
```
flavens, line 152: @tobiasschweizer when I run this test, this expect fails because the method is called multiple times in a random way (sometimes it is called 1, 3, 6, 7 ,8 ,9, etc...). I do not understand the logic behind the error message. Maybe I have to add something in the beforeEach?

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+            { query: 'new thing' }
+        ];
+        expect(localStorage.getItem('prevSearch')).toEqual(JSON.stringify(newPrevSearchArray));
+
+        // check the call of router.navigate with correct arguments
+        expect(mockRouter.navigate).toHaveBeenCalledWith(['/search/fulltext/new thing']);
+
+    });
+
+    // todo: fix error: Cannot read property 'detach' of undefined CDK OVERLAY
```
flavens, line 190: @kilchenmann @tobiasschweizer I get the following error: Cannot read property 'detach' of undefined. I suppose I have to define Overlay somewhere but I looked everywhere and tried several tutorials unsuccessfully.... Any idea?

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+
+        expect(testHostComponent.fulltextSearch.searchQuery).toEqual('purple thing');
+        expect(testHostComponent.fulltextSearch.projectIri).toEqual('http://rdfh.ch/projects/0801');
+        expect(testHostComponent.fulltextSearch.projectLabel).toEqual('anything');
+
+        expect(mockRouter.navigate).toHaveBeenCalledWith(['/search/fulltext/purple thing/http%3A%2F%2Frdfh.ch%2Fprojects%2F0801']);
+
+    });
+
+    // todo: fix error: Cannot read property 'detach' of undefined CDK OVERLAY
```
flavens, line 236: @kilchenmann @tobiasschweizer same comment sa above: I get the following error: Cannot read property 'detach' of undefined. I suppose I have to define Overlay somewhere but I looked everywhere and tried several tutorials unsuccessfully.... Any idea?

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+        testHostFixture.detectChanges();
+
+        const newPrevSearchArray: PrevSearchItem[] = [
+            { projectIri: 'http://rdfh.ch/projects/0803', projectLabel: 'incunabula', query: 'two things' },
+            { projectIri: 'http://rdfh.ch/projects/0803', projectLabel: 'incunabula', query: 'one thing' }
+        ];
+        expect(localStorage.getItem('prevSearch')).toEqual(JSON.stringify(newPrevSearchArray));
+    });
+
+    // todo: fix test - ERROR: Cannot read property 'indexOf' of null
```
flavens, line 326: @tobiasschweizer I get the following error: Cannot read property 'indexOf' of null. I don't understand. This test calls directly the method I want to check (l.339). Any idea?

projects/dsp-ui/src/lib/search/search-panel/search-panel.component.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    beforeEach(() => {
+        testHostFixture = TestBed.createComponent(TestHostComponent);
+        testHostComponent = testHostFixture.componentInstance;
+        testHostFixture.detectChanges();
+
+        expect(testHostComponent).toBeTruthy();
+    });
+
+    it('should create an instance', () => {
```
flavens, line 56: @tobiasschweizer I do not know which other tests I have to implement for this component because it is only a "combiner" (component to assemble the search bar with other components). And it uses mostly CDK OverlayRef which I do not know how to define or mock in a unit test....

tobiasschweizer:
@flavens I will look at your comment tomorrow.

tobiasschweizer:
@flavens I think the mock of `localStorage` is not called in the fulltext search spec. Does it work for you?

This could be an alternative approach without mocking the localStorage:

- in the "beforeEach", do the following: `localStorage.setItem('currentProject', JSON.stringify(MockProjects.mockProject().body.project));`
- then in the `it` "should get projects on init":  you can check for `testHostComponent.fulltextSearch.projectLabel`. This should be different depending on whether a `'currentProject'` was set or not. You could make different `describe` blocks where the `localStorage` contains different items.

tobiasschweizer:
@flavens Now I could see that the localStorage spy was called (when running the tests without the `fit`). Maybe I missed something.

tobiasschweizer:
@flavens I think you should move the `beforeEach` containing the `localStorage` spies up, so it is placed before the `beforeEach` where the component instantiation happens.

tobiasschweizer:
otherwise the spies definition happens **after** the component is created.

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
tobiasschweizer: When I run the test "'should do a search'", I get the following errors:

Screenshot 2020-06-23 at 12 52 59

I am not sure where the zone errors (XMLHTTPRequest) come from. The router or some other dependency that should be mocked? 

I think the `detach` problem comes from an uninitialised `this.overlayRef`. In this case, it is `undefined`.

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
tobiasschweizer: When I run the test "'should do a search'", I get the following errors:

Screenshot 2020-06-23 at 12 52 59

I am not sure where the zone errors (XMLHTTPRequest) come from. The router or some other dependency that should be mocked? 

I think the `detach` problem comes from an uninitialised `this.overlayRef`. In this case, it is `undefined`.

flavens: yes this is my thought too but I could not find the proper way to initiate OverlayRef...

tobiasschweizer: maybe you could mock it like the router?

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
tobiasschweizer: `prevSearch: PrevSearchItem[] = JSON.parse(localStorage.getItem('prevSearch'));` is set to `null` in case the item does not exist. I think you should handle the initialisation of  prevSearch differently and set it to an empty array if there is no previous search.

projects/dsp-ui/src/lib/search/search-panel/search-panel.component.spec.ts
tobiasschweizer: you can always mock the children in your spec file. The just need to have the same selector and the same `@input`s, `@Output`s, and public methods. However, you do not really have to implement them. There are like a mock.

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
flavens: I do not understand why here `prevSearch` is set to null. The test should remove one item of the search list so I need a list. The solution 1 works.

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
flavens: @tobiasschweizer I have tried multiple ideas and tutorials, nothing seems to work

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
tobiasschweizer: Independent of you test, `prevSearch: PrevSearchItem[] = JSON.parse(localStorage.getItem('prevSearch'));` is not good because the member `prevSearch` could be assigned `null`.

Rather you should do it in `ngOnInit`:

```

const prevSearchOption = JSON.parse(localStorage.getItem('prevSearch'));
if (prevSearchOption !== null) {
   this.prevSearch = prevSearchOption;
} else {
   this.prevSearch = [];
}
```

Like this, you always have an array which is compatible with `prevSearch: PrevSearchItem[]`.

We should think about using `strictNullChecks` (https://www.typescriptlang.org/docs/handbook/compiler-options.html), so this is enforced at compile time.

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
tobiasschweizer: Couldn't you just make an object with the methods you need and pass it to the `providers`? Otherwise we can look at this together.

tobiasschweizer:
@flavens I will take a look at this PR this afternoon.

projects/dsp-ui/src/lib/search/search-panel/search-panel.component.spec.ts
flavens: done in 02d9312

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
flavens: OverlayRef is part of the [CDK Overlay](https://material.angular.io/cdk/overlay/overview) from Angular Material. It is for me a box magic box and I have no idea how I should mock it or make an object of it with methods...

projects/dsp-ui/src/lib/search/fulltext-search/fulltext-search.component.spec.ts
flavens: prevSearch is always set to [] even if the localStorage has a key `prevSearch`

tobiasschweizer: But wouldn't it be set to `null` if there is no item fot he given key in the localStorage? `localStorage.getItem` returns `null` if the key does not exist, right?

My understanding is that this lead to ` Cannot read property 'indexOf' of null`

flavens: `prevSearch` is undefined and `prevSearchOption` is null despite the fact I set the localStorage in the beforeEach... Is there another setting to do?

tobiasschweizer: I think I would have to try to go through the code myself. Maybe then it gets clearer for me.

PR #118, opened 2020-07-02
AppInitService: Make members non static

tobiasschweizer:
I am trying to simplify the `AppInitService`.

I found out that providers can have [dependencies](https://medium.com/@gmurop/managing-dependencies-among-app-initializers-in-angular-652be4afce6f) (https://medium.com/@gmurop/managing-dependencies-among-app-initializers-in-angular-652be4afce6f), and those are then passed to the [factory method](https://angular.io/guide/dependency-injection-providers#factory-providers):

https://github.com/dasch-swiss/knora-ui-ng-lib/blob/873444569aac60113aee517e0cdfaa5628a78f8c/src/app/app.module.ts#L55-L73

I think there is no need for the `AppInitService` to have static members as it is instantiated by Angular. 

The problem is that whenever I try to use `@Inject(DspApiConfigToken) private dspApiConfig: KnoraApiConfig` in a constructor, I get an error because the dependency injection does not seem to work correctly.


kilchenmann:
We had to implement the AppInitService to load the configuration after the app was built. The application is built as a docker image and can be used with different configurations in different environments. This is the main reason for the complicated [setup](https://www.npmjs.com/package/@dasch-swiss/dsp-ui#setup).

> The problem is that whenever I try to use @Inject(DspApiConfigToken) private dspApiConfig: KnoraApiConfig in a constructor, I get an error because the dependency injection does not seem to work correctly.

We've never had this issue before but I will have a look at it...

tobiasschweizer:
check https://github.com/dasch-swiss/knora-ui/blob/master/projects/knora/core/src/lib/core.module.ts

tobiasschweizer:
@kilchenmann @subotic I think I found a way to get the config without changing `main.ts`, see https://github.com/dasch-swiss/knora-ui-ng-lib/pull/118/commits/9a72c3361dceb789a118ee258c39c25a094d8705

Since we use the `APP_INITIALIZER` token, we just return a Promise that is resolved once the configuration is assigned to `AppInitService`s member `dspApiConfig`.

> The provided function are injected at application startup and executed during app initialization. If any of these functions returns a Promise, initialization does not complete until the Promise is resolved.

https://angular.io/api/core/APP_INITIALIZER

tobiasschweizer:
I will try to handle the case where getting the config would fail (rejected Promise).

kilchenmann:
> I think I found a way to get the config without changing main.ts, see 9a72c33

Will it work in a dockerized app too? Can you make an example? I'm asking because we had a similar solution first but it failed on the docker side with the configuration from outside.

tobiasschweizer:
> Will it work in a dockerized app too? Can you make an example? I'm asking because we had a similar solution first but it failed on the docker side with the configuration from outside.

How can I find out whether it works in docker? Is there a way to build the app in docker and run it?

The config file is loaded asynchronously like in the previous solution, only that this way, `main.ts` does not have to be changed. `AppInitService`s `Init` method returns a `Promise` that is resolved once the config has fetched read and assigned. Angular wait until the Promise is resolved.

tobiasschweizer:
@kilchenmann What about https://github.com/dasch-swiss/knora-api-js-lib/blob/master/test-framework/Dockerfile?

kilchenmann:
> What about https://github.com/dasch-swiss/knora-api-js-lib/blob/master/test-framework/Dockerfile?

It's to build the js-lib test-framework app as a docker container. But I have no idea for what it should be used

tobiasschweizer:
I managed to build the lib and app in docker, now I am trying to tun it ...

kilchenmann:
> How can I find out whether it works in docker? Is there a way to build the app in docker and run it?

You can try to build the dsp-ui-lib test-app as a docker container locally. And run it in your docker environment by using the configuration from <https://github.com/dasch-swiss/ops-deploy/tree/master/knora-stack>

kilchenmann:
> I managed to build the lib and app in docker, now I am trying to run it ...

Sorry I didn't saw your new comment

tobiasschweizer:
@kilchenmann https://github.com/dasch-swiss/knora-ui-ng-lib/pull/118/commits/d0a8df312cd97256f158458034e0e9ece0faccc0 sets up the build process config with docker:

- create a config file on your local file system, for example `/Desktop/config/config.prod.json`
- `docker build --tag app:0.1 .`
- `docker run -v ~/Desktop/config/config.prod.json:/public/config/config.prod.json --publish 4200:4200 --detach --name bb1 app:0.1`

This works for me and shows me:

Screenshot 2020-07-06 at 14 22 51

The config file in my app looks like this:

```json
{
  "apiProtocol": "http",
  "apiHost": "0.0.0.0",
  "apiPort": 3333,
  "apiPath": "",
  "jsonWebToken": "",
  "logErrors": false
}
```

and the file `/Desktop/config/config.prod.json` like this:

```json
{  
"apiProtocol": "http",
  "apiHost": "0.0.0.0",
  "apiPort": 3333,
  "apiPath": "",
  "jsonWebToken": "gagagaga",
  "logErrors": false
}
```

kilchenmann:
Cool. Thanks

src/app/app.module.ts
```diff
@@ -30,45 +30,47 @@ export function initializeApp(appInitService: AppInitService) {
...
+      deps: [AppInitService]
+    },
+    {
+      provide: DspApiConnectionToken,
+      useFactory: (appInitService: AppInitService) => new KnoraApiConnection(appInitService.dspApiConfig),
+      deps: [AppInitService]
+    }
+  ],
+  bootstrap: [AppComponent]
+
```
kilchenmann, line 73: Indent should be 4 spaces in typescript files.

kilchenmann:
It worked. 
@tobiasschweizer Can you please update the README in the `projects` folder?

tobiasschweizer:
> It worked.
> @tobiasschweizer Can you please update the README in the `projects` folder?

Great!

Yes, I will. So we can remove the adaption of `main.ts`?

src/app/app.module.ts
tobiasschweizer: fixed in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/118/commits/262c9545e62ba2b3d947f5b3b63b96bb8b5b3daf

tobiasschweizer:
Could we move the `AppInitService` to the core module?

tobiasschweizer:
@kilchenmann I moved the `AppInitService` to the lib so that users of the lib do not have to reimplement it. Will change the README as requested.

kilchenmann:
> I moved the AppInitService to the lib so that users of the lib do not have to reimplement it.

Good idea

tobiasschweizer:
> It worked.
> @tobiasschweizer Can you please update the README in the `projects` folder?

done in https://github.com/dasch-swiss/knora-ui-ng-lib/pull/118/commits/453524bd2158eb6ecc3dc553133d505bd4b58737

tobiasschweizer:
@kilchenmann @mdelez I would like to merge this PR into wip/DSP-238-session-service-migration.

PR #121, opened 2020-07-08
Flaky E2E Tests

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-457

The E2E tests for the advanced search failed from time to time because async operations were no taken into account properly. For example, when an ontology has been selected, the UI has to get the ontology or ontologies from Knora. This could take some time. Unfortunately, the Harnesses in E2E tests [do not implement the necessary functionality yet](https://github.com/angular/components/issues/18375). 

As a solution, I use `browser.ExpectedConditions`: information fetched from Knora is expected to be displayed in the GUI within a defined time limit. Once the condition is met, the tests continues and further assertions can be checked safely.

### DIFF
--- a/e2e/protractor.conf.js
+++ b/e2e/protractor.conf.js
@@ -8,7 +8,7 @@ const {SpecReporter} = require('jasmine-spec-reporter');
  * @type { import("protractor").Config }
  */
 exports.config = {
-    allScriptsTimeout: 11000,
+    allScriptsTimeout: 22000,
     specs: [
         './src/**/*.e2e-spec.ts',
     ],
--- a/e2e/src/app.e2e-spec.ts
+++ b/e2e/src/app.e2e-spec.ts
@@ -6,6 +6,7 @@ import { MatAutocompleteHarness } from '@angular/material/autocomplete/testing';
 
 describe('Test App', () => {
     let page: AppPage;
+    const timeout = 6000;
 
     beforeEach(() => {
         page = new AppPage();
@@ -25,7 +26,7 @@ describe('Test App', () => {
         it('should display an integer value', async () => {
             await page.navigateTo('modify');
 
-            const valueEleComp: WebElement = await page.getComponentBySelector('dsp-int-value');
+            const valueEleComp: WebElement = await page.getComponentBySelector('dsp-int-value', timeout);
 
             const intValEleField = await page.getReadValueFieldFromValueComponent(valueEleComp);
             expect(await intValEleField.getText()).toEqual('1');
@@ -36,7 +37,7 @@ describe('Test App', () => {
 
             await page.navigateTo('modify');
 
-            const valueEleComp: WebElement = await page.getComponentBySelector('dsp-int-value');
+            const valueEleComp: WebElement = await page.getComponentBySelector('dsp-int-value', timeout);
 
             const displayEditComp: WebElement = await page.getDisplayEditComponentFromValueComponent(valueEleComp);
 
@@ -56,7 +57,7 @@ describe('Test App', () => {
 
             const EC = browser.ExpectedConditions;
 
-            await browser.wait(EC.presenceOf(element(by.css('.rm-value'))), 3000,
+            await browser.wait(EC.presenceOf(element(by.css('.rm-value'))), timeout,
                 'Wait for read value to be visible.');
 
             const readEle = await page.getReadValueFieldFromValueComponent(valueEleComp);
@@ -122,7 +123,7 @@ describe('Test App', () => {
             await selectOntos.clickOptions({ text: 'The anything ontology'});
 
             // check for the async response from Knora: anything and knora-api ontology
-            await browser.wait(EC.presenceOf(element(by.css('.select-resource-class'))), 3000,
+            await browser.wait(EC.presenceOf(element(by.css('.select-resource-class'))), timeout,
                 'Wait for resource class options to be visible.');
 
             expect(await submitButton.isDisabled()).toBe(true);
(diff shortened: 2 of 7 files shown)


flavens:
@tobiasschweizer Before giving my review, I tested it on my machine. I got 2 error messages (not about the advanced search e2e tests):

Screenshot 2020-07-09 at 08 24 41

Did I miss a step?

tobiasschweizer:
Hm, I have never seen this error message before. Have you reloaded the data before?

@mdelez Is this the message you got sometimes?

Maybe another timing issue.

flavens:
> Hm, I have never seen this error message before. Have you reloaded the data before?

I reloaded the data 3 days ago when I switched from dsp-api rc5 to rc4 but I have reviewed many PRs and worked on my PR expert search since then.

tobiasschweizer:
@flavens Maybe there is something about timing I haven't properly understood yet. I general, I understood that the result of an operation involving Knora can take some time and therefore `browser.wait` is needed. For example when a value is updated and read back from Knora.

Maybe the integer value hasn't been read back from Knora yet, so I should use another `browser.wait` with a timeout.

tobiasschweizer:
> > Hm, I have never seen this error message before. Have you reloaded the data before?
> 
> I reloaded the data 3 days ago when I switched from dsp-api rc5 to rc4 but I have reviewed many PRs and worked on my PR expert search since then.

Before running the E2E tests, you should always reload the data and restart Knora because the E2E test changes the data. So when you run them again without reloading the data, it will fail. The test expects the integer value to be 2 and changes it to 3.

flavens:
> Before running the E2E tests, you should always reload the data and restart Knora because the E2E test changes the data. So when you run them again without reloading the data, it will fail. The test expects the integer value to be 2 and changes it to 3.

I reloaded the data. I am using dsp-api rc4. Still the exact same test fails.

tobiasschweizer:
> > Before running the E2E tests, you should always reload the data and restart Knora because the E2E test changes the data. So when you run them again without reloading the data, it will fail. The test expects the integer value to be 2 and changes it to 3.
> 
> I reloaded the data. I am using dsp-api rc4. Still the exact same test fails.

Ok, maybe a timing issue because the int value has to be fetched from Knora first. I will add a `browser.wait` with a timeout.

mdelez:
> Hm, I have never seen this error message before. Have you reloaded the data before?
> 
> @mdelez Is this the message you got sometimes?
> 
> Maybe another timing issue.

Yeah, these are the errors I was getting. Reloading the data fixed it but like you mentioned below, you need to reload the data every time you want to run the tests which really isn't feasible.

tobiasschweizer:
> you need to reload the data every time you want to run the tests which really isn't feasible.

Since the e2e tests changes the data, there is no other way to achieve this. The e2e tests rely on certain expectations, and only reloading the data ensures that these expectations can be met.

However, I still think there might be a timing problem. I will work on that.

tobiasschweizer:
> > Before running the E2E tests, you should always reload the data and restart Knora because the E2E test changes the data. So when you run them again without reloading the data, it will fail. The test expects the integer value to be 2 and changes it to 3.
> 
> I reloaded the data. I am using dsp-api rc4. Still the exact same test fails.

I added another `browser.wait` before getting `dsp-int-value`. I think your problem could be a timing issue since the element only exists when the value has been read back from Knora and is displayed in the template.

Could you try again? @mdelez Could you also give it a try?

flavens:
> Could you try again? @mdelez Could you also give it a try?

@tobiasschweizer e2e tests failed again for me..

DSP-API 13.0.0-rc4 (data reloaded)

I tried `npm run e2e` 3 times:

Test 1.
Screenshot 2020-07-09 at 18 04 46

Test 2.
Screenshot 2020-07-09 at 18 12 59

Test 3.
Screenshot 2020-07-09 at 18 15 40

tobiasschweizer:
I think this means that the value is not displayed within 3 seconds. Could you access the modify page in the playground and see how long it takes until the integer value is displayed?

flavens:
> I think this means that the value is not displayed within 3 seconds. Could you access the modify page in the playground and see how long it takes until the integer value is displayed?

Visually, all components are displayed quickly and at the same time. According the network monitoring, I got this timing or less to display the resource (I do not know if it is possible to the timing only for the integer value component):

Screenshot 2020-07-10 at 09 42 05

mdelez:
> Could you try again? @mdelez Could you also give it a try?

DSP-API 13.0.0-rc4 (data reloaded)

I also tried npm run e2e 3 times:

Test 1.
Screen Shot 2020-07-10 at 09 41 57

Test 2.
Screen Shot 2020-07-10 at 09 42 57

Test 3.
Screen Shot 2020-07-10 at 09 43 53

Test 4. (data reloaded)
Screen Shot 2020-07-10 at 10 00 41

tobiasschweizer:
@flavens @mdelez Could we set up a meeting next week so you could share your screen with me when running the E2E tests?

flavens:
> @flavens @mdelez Could we set up a meeting next week so you could share your screen with me when running the E2E tests?

Yes sure. Monday at 10am? Or 2pm?

tobiasschweizer:
10 a.m is fine. Shall I send an invite?

flavens:
Running e2e tests for the 6th times (xit "should edit an integer value"):

Screenshot 2020-07-13 at 10 31 17

tobiasschweizer:
This PR provides some improvements (assertions with a timeout) but it does not solve the problem of the flaky test properly. I will do this in https://dasch.myjetbrains.com/youtrack/issue/DSP-474.

tobiasschweizer:
@flavens I made a new DSP to implement my own Harnesses, see https://dasch.myjetbrains.com/youtrack/issue/DSP-474.

Is this PR good to go?

flavens:
> Is this PR good to go?

Yes!

PR #124, opened 2020-07-13
Create custom Harness for select ontology comp.

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-474


tobiasschweizer:
https://github.com/dasch-swiss/dsp-ui-lib/pull/124/commits/5546faad472d2d73dbf684c60ab7ce64db3a3a32 adds the path config to `protractor.conf.js` so that the lib can be accessed (imported) in the e2e spec. I got the solution from https://github.com/nrwl/nx/issues/263#issuecomment-485889032.

However, there is still a problem with the `window` object. When importing from `@dasch-swiss/dsp-ui`, `jsonld` is processed by `ts-node`. This throws an error because the `window` object is accessed, but not set in the `node` env:

> E/launcher - Error: ReferenceError: window is not defined

PR #125, opened 2020-07-14
Login Form Migration

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-480


projects/dsp-ui/src/lib/action/components/login-form/login-form.component.html
```diff
@@ -0,0 +1,69 @@
+<div class="login-container" *ngIf="!session">
+
+    <h2 class="login-form-title mat-title">{{formLabel.title}}</h2>
+
+    <!-- This is the login form -->
+    <form [formGroup]="form" class="login-form">
+
+        <!-- Username -->
+        <mat-form-field class="full-width login-field">
+            <!-- <mat-icon matPrefix>person</mat-icon> -->
```
flavens, line 10: If we are not using the icon, you should remove the line. Same in line 24.

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+import { of } from 'rxjs';
+import { AjaxResponse } from 'rxjs/ajax';
+import { DspApiConnectionToken, Session, SessionService } from '../../../core';
+import { LoginFormComponent } from './login-form.component';
+
+/**
+ * Test host component to simulate login-form component.
+ */
+@Component({
+    template: `<dsp-login-form #loginForm></dsp-login-form>`
```
flavens, line 23: In the test host component, you should register to the @Output status. Then you can trigger the emission and see if it emits the correct status.

flavens: like in:
https://github.com/dasch-swiss/dsp-ui-lib/blob/b3b1fee1ecc97085f3bdb199c0352c36825b206a/projects/dsp-ui/src/lib/search/advanced-search/select-ontology/select-ontology.component.spec.ts#L14-L44

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.html
kilchenmann: This is the question: Should we display nice icons in the form field or not?

kilchenmann: Screenshot 2020-07-17 at 06 09 09

VS

Screenshot 2020-07-17 at 06 09 19

projects/dsp-ui/src/lib/action/action.module.ts
```diff
@@ -15,6 +15,9 @@ import { SortByPipe } from './pipes/array-transformation/sort-by.pipe';
 import { FormattedBooleanPipe } from './pipes/formatting/formatted-boolean.pipe';
 import { KnoraDatePipe } from './pipes/formatting/knoradate.pipe';
 import { TruncatePipe } from './pipes/string-transformation/truncate.pipe';
+import { LoginFormComponent } from './components/login-form/login-form.component';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { MatInputModule } from '@angular/material/input';
```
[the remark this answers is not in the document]
kilchenmann: This is also what I saw in other files. I think, we have to refactor all files by cleaning up the imports. Good point!

@mdelez You can use the "Organize Imports" command in vsc: `shift` + `opt` + `o`

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.html
kilchenmann: I prefer the version without icons. So yes, I agree. @mdelez you can delete the commented mat-icons.

projects/dsp-ui/src/lib/action/action.module.ts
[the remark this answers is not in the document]
mdelez: done in 5ff6925

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.spec.ts
mdelez: I attempted this in 5ff6925. However, in the callback, the status is emitted within a setTimeout function which makes it so that my test fails because it isn't emitted in time for the assertion.

@kilchenmann does this need to be in a setTimeout function?
https://github.com/dasch-swiss/dsp-ui-lib/blob/5ff6925f133de3b7ccb027fa10c8739c7b5623ac/projects/dsp-ui/src/lib/action/components/login-form/login-form.component.ts#L139-L142

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.html
mdelez: I like the icons. Is this something we can set up so the user can choose to display it or not?

kilchenmann: > Is this something we can set up so the user can choose to display it or not?

Why not? Can you add an `@Input` "icons" (or something similar) of type boolean (default `false`) to show them or not...

mdelez:
@kilchenmann  I also find this a bit confusing
The status Output claims that it will emit false after the logout process
https://github.com/dasch-swiss/dsp-ui-lib/blob/5ff6925f133de3b7ccb027fa10c8739c7b5623ac/projects/dsp-ui/src/lib/action/components/login-form/login-form.component.ts#L28-L34

but then in the logout process we have
https://github.com/dasch-swiss/dsp-ui-lib/blob/5ff6925f133de3b7ccb027fa10c8739c7b5623ac/projects/dsp-ui/src/lib/action/components/login-form/login-form.component.ts#L167-L175

The equality check in the emission will return true because the status within the body is 0. This should return false, right? Also, does there need to be an equality check? Can we just emit false?

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.html
mdelez: added in 32e144c

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.spec.ts
mdelez: refactored in 32e144c

mdelez:
> The equality check in the emission will return true because the status within the body is 0. This should return false, right? Also, does there need to be an equality check? Can we just emit false?

I changed it to emit false in 32e144c

kilchenmann:
> Can we just emit false?

In case of successful logout we return `true`. In case of an error, return `false`. But not sure if it's the best idea?

mdelez:
> > Can we just emit false?
> 
> In case of successful logout we return `true`. In case of an error, return `false`. But not sure if it's the best idea?

The comment says it will return false after the logout process
```Emits true when the login process was successful and false in case of error on login or false after logout process```

but I'm a bit confused on the purpose of this status Output. I think I'm getting confused with the status Input that is used in the template for the Progress Indicator. These two are unrelated, correct?

I'm struggling to understand what the status Output is used for. Is it supposed to be used to tell if the user is logged in or out or is is supposed to be used to tell if an operation was successful or not?

**Comment by kilchenmann** [comment]
> The comment says it will return false after the logout process
> `Emits true when the login process was successful and false in case of error on login or false after logout process`
> 
> but I'm a bit confused on the purpose of this status Output. I think I'm getting confused with the status Input that is used in the template for the Progress Indicator. These two are unrelated, correct?
> 
> I'm struggling to understand what the status Output is used for. Is it supposed to be used to tell if the user is logged in or out or is it supposed to be used to tell if an operation was successful or not?
> 
> Also, nothing is emitted in either of the two error cases. Maybe we should look at this together to decide on a way this should be implemented?

I understand and agree that it's confusing. But we can decide now, how we should handle the login/logout returns. When I want to use the login form in my app, I'll need the output (true for successful login) to do something in the app after login e.g. redirect to user's profile page. The same should be for logout. I need a response to know what I have to do. 
We should discuss it now and implement the best solution.

File: projects/dsp-ui/src/lib/action/components/login-form/login-form.component.ts [hunk 4]
```diff
@@ -0,0 +1,200 @@
...
+        name: 'Username',
+        pw: 'Password',
+        submit: 'Login',
+        retry: 'Retry',
+        logout: 'LOGOUT',
+        remember: 'Remember me',
+        forgot_pw: 'Forgot password?',
+        error: {
+            failed: 'Password or username is wrong',
+            server: 'There\'s an error with the server connection. Try it again later or inform the Knora Team'
```
flavens, line 81: @mdelez @kilchenmann small detail: the final dot is missing, and should we now use "Dasch Team" instead of "Knora Team"?

mdelez:
> LGTM!
> See my previous comment about Knora Team

done in 7b6b1fa

projects/dsp-ui/src/lib/action/action.module.ts
```diff
@@ -28,14 +31,18 @@ import { TruncatePipe } from './pipes/string-transformation/truncate.pipe';
     GndDirective,
     ProgressIndicatorComponent,
     MessageComponent,
-    SortButtonComponent
+    SortButtonComponent,
+    LoginFormComponent
   ],
   imports: [
     CommonModule,
     MatIconModule,
```
kilchenmann, line 41: Add `MatButtonModule` here

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.ts
kilchenmann: Yes, it should be "DaSCH Team."

projects/dsp-ui/src/lib/action/action.module.ts
kilchenmann: I was a bit slow. Already done 👍

PR #126, opened 2020-07-14
Simple grid of resources

flavens:
closes [DSP-481](https://dasch.myjetbrains.com/youtrack/issue/DSP-481)


projects/dsp-ui/src/lib/viewer/views/list-view/resource-grid/resource-grid.component.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        item.dispatchEvent(new Event('click'));
+
+        spyOn(testHostComponent, 'openResource').call('http://rdfh.ch/0001/H6gBWUuJSuuO-CilHV8kQw');
+        expect(testHostComponent.openResource).toHaveBeenCalled();
+        expect(testHostComponent.openResource).toHaveBeenCalledTimes(1);
+
+        // console.log('resIri', testHostComponent.resources.resources[0].id);
+        // console.log('resIri', testHostComponent.resIri);
+        expect(testHostComponent.resources.resources[0].id).toEqual('http://rdfh.ch/0001/H6gBWUuJSuuO-CilHV8kQw');
+        // expect(testHostComponent.resIri).toEqual('http://rdfh.ch/0001/H6gBWUuJSuuO-CilHV8kQw');
```
flavens, line 95: @kilchenmann I compared this test with the one in resource-list and for some reasons resIri is undefined... The only difference in this test is `nativeElement.querySelector('mat-card')`. Do you have an idea?

kilchenmann: Let me check that...

kilchenmann: Why do you use an additional `<div>`? 

This `<div *ngFor="let res of resources.resources" (click)="resourceSelected.emit(res.id)" class="link res-grid-item">` should be added to `<mat-card class="grid-card">`. No `div` is needed. And you test the link on `mat-card`, so it should work then.

flavens: My mistake, I tried different html construction and I kept the div. I moved the `*ngFor` in `mat-card` and it works well now.

projects/dsp-ui/src/lib/viewer/views/list-view/resource-grid/resource-grid.component.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    it('should open first resource', () => {
+        // trigger the click
+        const nativeElement = testHostFixture.nativeElement;
+        const item = nativeElement.querySelector('mat-card');
+        item.dispatchEvent(new Event('click'));
+
+        spyOn(testHostComponent, 'openResource').call('http://rdfh.ch/0001/H6gBWUuJSuuO-CilHV8kQw');
+        expect(testHostComponent.openResource).toHaveBeenCalled();
```
mdelez, line 89: this assertion isn't needed as you are also checking it was called below

kilchenmann: Yes, I think it's the same we already had in resource-list. I'll update both in the next PR

PR #132, opened 2020-07-20
Migrate Stringify String Literal Pipe

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-504

### DIFF
--- a/projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts
@@ -1,5 +1,6 @@
 import { Pipe, PipeTransform } from '@angular/core';
 import { StringLiteral } from '@dasch-swiss/dsp-js';
+import { SessionService } from '../../../core';
 
 /**
  * This pipe stringifies an array of StringLiterals.
@@ -14,6 +15,8 @@ import { StringLiteral } from '@dasch-swiss/dsp-js';
   name: 'dspStringifyStringLiteral'
 })
 export class StringifyStringLiteralPipe implements PipeTransform {
+
+    constructor(private _sessionService: SessionService) { }
 
     transform(value: StringLiteral[], args?: string): string {
         let stringified = '';
@@ -38,9 +41,9 @@ export class StringifyStringLiteralPipe implements PipeTransform {
             // show only one value, depending on default language
             // the language is defined in user profile if a user is logged-in
             // otherwise it takes the language from browser
-            if (localStorage.getItem('session') !== null) {
+            if (this._sessionService.getSession() !== null) {
                 // get language from the logged-in user profile data
-                language = JSON.parse(localStorage.getItem('session')).user.lang;
+                language = this._sessionService.getSession().user.lang;
             } else {
                 // get default language from browser
                 language = navigator.language.substr(0, 2);
--- a/projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts
@@ -1,0 +1,61 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { StringLiteral } from '@dasch-swiss/dsp-js';
+
+/**
+ * This pipe stringifies an array of StringLiterals.
+ * With the parameter 'all', the pipe concats all values and appends the corresponding language in brackets.
+ *
+ * Otherwise the pipe displays the value corresponding to the default language which
+ * comes from the user profile if a user is logged in or from the browser if the user is not logged in.
+ *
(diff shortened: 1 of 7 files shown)


mdelez:
Two things to note @kilchenmann:
1. I think we should somehow provide a warning if you supply the pipe with an invalid parameter. For example:
```<p>{{labels | dspStringifyStringLiteral:'invalidParameter'}}</p>```
will just return a single value with no indication of why the parameter provided is not working. We could use the union type or enum to solve this and if the parameter does not match, we give a warning. I admit this is kind of a bit much to go through when there is only one valid parameter the user can provide to the pipe, but I think it would make using the pipe clearer.

2. I don't know of a way to test the case where the pipe gets the language of the browser in the event that the user is not logged in. This is dependent on the language of the browser/computer from which the tests are being run on.

File: projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+                stringified += delimiter + sl.value + ' (' + sl.language + ')';
+
+                i++;
+            }
+            return stringified;
+        } else {
+            // show only one value, depending on default language
+            // the language is defined in user profile if a user is logged-in
+            // otherwise it takes the language from browser
+            if (localStorage.getItem('session') !== null) {
```
kilchenmann, line 41: We shouldn't use localStorage.getItem direct to get the `session` anymore. Please use the session service here.

projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts
```diff
@@ -0,0 +1,61 @@
...
+                // get language from the logged-in user profile data
+                language = JSON.parse(localStorage.getItem('session')).user.lang;
```
kilchenmann, line 43: Same as above...

kilchenmann:
> I think we should somehow provide a warning if you supply the pipe with an invalid parameter. For example:
<p>{{labels | dspStringifyStringLiteral:'invalidParameter'}}</p>
will just return a single value with no indication of why the parameter provided is not working. We could use the union type or enum to solve this and if the parameter does not match, we give a warning. I admit this is kind of a bit much to go through when there is only one valid parameter the user can provide to the pipe, but I think it would make using the pipe clearer.

Yes, I agree. I think the primary idea of `args` was to expand later and to give not only `all` but a specific language. 
`<p>{{labels | dspStringifyStringLiteral:'de'}}</p>` 
But what should we do, if this specific language doesn't exist or doesn't have a value? You can change the `args` into a boolean value type e.g. `allLang = true` will display all, otherwise it takes the browser's or user's default language. But we should keep in mind that an app can have a language switch button to change the language of the interface which will include the StringLiteral values. So, I prefer to extend the `args`:
`transform(value: StringLiteral[], args?: 'all' | 'de' | 'en' | 'fr' | 'it'): string {`

The default one will be the one from browser or from user.

projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts
mdelez: fixed 0055441

projects/dsp-ui/src/lib/action/pipes/string-transformation/stringify-string-literal.pipe.ts
mdelez: fixed in 0055441

mdelez:
> Yes, I agree. I think the primary idea of `args` was to expand later and to give not only `all` but a specific language.
> `<p>{{labels | dspStringifyStringLiteral:'de'}}</p>`
> But what should we do, if this specific language doesn't exist or doesn't have a value? You can change the `args` into a boolean value type e.g. `allLang = true` will display all, otherwise it takes the browser's or user's default language. But we should keep in mind that an app can have a language switch button to change the language of the interface which will include the StringLiteral values. So, I prefer to extend the `args`:
> `transform(value: StringLiteral[], args?: 'all' | 'de' | 'en' | 'fr' | 'it'): string {`
> 
> The default one will be the one from browser or from user.

The language switching will be tricky. Do we expect users to be able to do this freely without it affecting the language in their session or will this language switch change the value of the language in the session object? What if the user is not logged in, are they allowed to switch the language? If so, this should probably be persistent somehow even if the user is a "guest".

kilchenmann:
> Do we expect users to be able to do this freely without it affecting the language in their session or will this language switch change the value of the language in the session object? What if the user is not logged in, are they allowed to switch the language? If so, this should probably be persistent somehow even if the user is a "guest".

It's just an idea in the app, but we have to keep in mind to change the stringliteral language quite fast. We can keep this pipe as it is right now. We have to refactor the lib one day anyway to support multilanguage interfaces and to make it more flexible. It would be too much now anyway.

PR #139, opened 2020-07-29
Create New Value V2

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-446


mdelez:
@tobiasschweizer why do the decimal and URI properties have a guiElement of SimpleText?

Screen Shot 2020-07-29 at 13 52 25

tobiasschweizer:
> @tobiasschweizer why do the decimal and URI properties have a guiElement of SimpleText?
> 
> Screen Shot 2020-07-29 at 13 52 25

I think the reason is that in the old GUI we had a set of GUI elements. Some elements like an input box could be used for different types of data.
This has now be replaced by Angular Material, and we defined our own components per data type. I think we will need the information in `salsah:guiElement` or `salsah:guiAttributes` in certain cases where the value type alone is not enough. Text values are one example (with and without markup).

projects/dsp-ui/src/lib/viewer/services/value-operation-event.service.ts
```diff
@@ -1,5 +1,6 @@
 import { Subject, Subscription } from 'rxjs';
 import { filter, map } from 'rxjs/operators';
+import { ReadValue } from '@dasch-swiss/dsp-js';
```
flavens, line 3: VSC warning: Import sources within a group must be alphabetized.

mdelez:
> It looks good except for one thing:
> when you enter an incorrect value in the input, the save button is still enabled and if you press the save button, you get the following error message in the console and the save + cancel buttons are gone.
> Screenshot 2020-08-03 at 14 10 01
> I suggest to disable the save button when there is an error.

Good catch! This is a tricky one to fix since in the display-edit component form we can use the form's validity to disable this save button. However in the add-value component, the form is not initialized until after the view is initialized. I'll have to think of a way around this 🤔

projects/dsp-ui/src/lib/viewer/services/value-operation-event.service.ts
mdelez: fixed in 97213e5

mdelez:
> It looks good except for one thing:
> when you enter an incorrect value in the input, the save button is still enabled and if you press the save button, you get the following error message in the console and the save + cancel buttons are gone.
> Screenshot 2020-08-03 at 14 10 01
> I suggest to disable the save button when there is an error.

@flavens I've found a not-so-elegant way to fix this error by using setTimeout. I know it's frowned upon but it's done within synchronous logic so I don't think it should ever cause an issue. What do you think? Do you have any other ideas?

https://github.com/dasch-swiss/dsp-ui-lib/blob/24cae6e9dc279442ed972a5dbb6b3bd0959599ac/projects/dsp-ui/src/lib/viewer/operations/add-value/add-value.component.ts#L72-L77

flavens:
> @flavens I've found a not-so-elegant way to fix this error by using setTimeout. I know it's frowned upon but it's done within synchronous logic so I don't think it should ever cause an issue. What do you think? Do you have any other ideas?

Good, thanks. I do not have another idea at the moment. You can request again.

PR #134, opened 2020-07-27
Add CKeditor

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-467

Warning when building the app:
> Warning: Entry point '@ckeditor/ckeditor5-angular' contains deep imports into 

see https://github.com/ckeditor/ckeditor5-angular/issues/181

`TextValueAsXMLComponent` integrates the Angular lib for CKEditor with a custom build of CKEditor 5.
Some transformation of the markup created by CKEditor is required before sending it to Knora. I added this transformation to the app's configuration so the component is more flexible. By default, `TextValueAsXMLComponent` assumes the [standard mapping](https://docs.knora.org/03-apis/api-v1/xml-to-standoff-mapping/).

The playground page allows playing around with the markup, logging it, and submitting it to Knora.

Note that `TextValueAsXMLComponent` is not [integrated](https://github.com/dasch-swiss/dsp-ui-lib/pull/134#issuecomment-686958591) yet with `DisplayEditComponent`.

PR #140, opened 2020-07-29
Warning When Running Lib's Tests

kilchenmann:
https://dasch.myjetbrains.com/youtrack/issue/DSP-521

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
@@ -2,7 +2,6 @@ import { Component, Input, OnInit, ViewChild } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
-import { MatListModule } from '@angular/material/list';
 import { MatPaginatorModule } from '@angular/material/paginator';
 import { CountQueryResponse, MockResource, ReadResourceSequence, SearchEndpointV2 } from '@dasch-swiss/dsp-js';
 import { IFulltextSearchParams } from '@dasch-swiss/dsp-js/src/api/v2/search/search-endpoint-v2';
--- a/projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
@@ -99,7 +99,6 @@ describe('ListViewComponent', () => {
             imports: [
                 MatButtonModule,
                 MatIconModule,
-                MatListModule,
                 MatPaginatorModule
             ],
             providers: [
--- a/projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
@@ -1,29 +1,20 @@
-import { Component, Input, OnInit, ViewChild } from '@angular/core';
+import { Component, OnInit, ViewChild } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
+import { MatListModule } from '@angular/material/list';
 import { MatPaginatorModule } from '@angular/material/paginator';
 import { CountQueryResponse, MockResource, ReadResourceSequence, SearchEndpointV2 } from '@dasch-swiss/dsp-js';
 import { IFulltextSearchParams } from '@dasch-swiss/dsp-js/src/api/v2/search/search-endpoint-v2';
 import { of } from 'rxjs';
+import { DspActionModule } from '../../../action';
 import { DspApiConnectionToken } from '../../../core';
 import { ListViewComponent, SearchParams } from './list-view.component';
+import { ResourceGridComponent } from './resource-grid/resource-grid.component';
+import { ResourceListComponent } from './resource-list/resource-list.component';
 
 /**
- * Test host component to simulate child component, here resource-list.
- */
-@Component({
-    selector: `dsp-list-view`,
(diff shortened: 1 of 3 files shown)


projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
```diff
@@ -78,11 +69,15 @@ describe('ListViewComponent', () => {
         TestBed.configureTestingModule({
             declarations: [
                 ListViewComponent,
+                ResourceListComponent,
```
tobiasschweizer, line 72: You shouldn't use the actual components but mock them like here https://github.com/dasch-swiss/dsp-ui-lib/pull/134/commits/9fb5841c9c62555032859c13819d5252ca4d4fac

The idea is that you provide a mocked component with the same selector and the same inputs and outputs as the real component.

tobiasschweizer: so I think in your case you have to mock the `ResourceListComponent` and the `ResourceGridComponent`.

The mocks can have an empty template.

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
```diff
@@ -78,11 +69,15 @@ describe('ListViewComponent', () => {
...
+                ResourceGridComponent,
                 TestParentComponent
             ],
             imports: [
+                DspActionModule,
```
tobiasschweizer, line 77: Could you mock the things you need from the DSPActionModule?

In general, importing our own code in unit tests should be avoided (however, sometimes there is a good reason for not using a mock).

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
```diff
@@ -1,29 +1,20 @@
...
+import { DspActionModule } from '../../../action';
 import { DspApiConnectionToken } from '../../../core';
 import { ListViewComponent, SearchParams } from './list-view.component';
+import { ResourceGridComponent } from './resource-grid/resource-grid.component';
+import { ResourceListComponent } from './resource-list/resource-list.component';
 
 /**
- * Test host component to simulate child component, here resource-list.
- */
-@Component({
```
tobiasschweizer, line 15: if this is the spec for dsp-list-view, the you only have to mock its children and parent (if necessary)

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
kilchenmann: I see, this is what we had (but probably not correct). I will try to undo...

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
kilchenmann: Done in d4c83b0

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
kilchenmann: Done in d4c83b0

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
kilchenmann: Now I understand. Thanks for explanation.
Resolved in d4c83b0

projects/dsp-ui/src/lib/viewer/views/list-view/resource-grid/resource-grid.component.spec.ts
```diff
@@ -1,26 +1,33 @@
...
-class TestResourceGridComponent {
+class TestProgressIndicatorComponent {
 
-    @Input() resources: ReadResourceSequence;
+}
 
+/**
+ * Mocked truncate pipe from action module.
+ */
+@Pipe({name: 'dspTruncate'})
```
tobiasschweizer, line 21: I think that is a good solution!

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.spec.ts
```diff
@@ -78,7 +91,9 @@ describe('ListViewComponent', () => {
         TestBed.configureTestingModule({
             declarations: [
                 ListViewComponent,
-                TestParentComponent
+                TestParentComponent,
```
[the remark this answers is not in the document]
kilchenmann: It worked without. Should I add it?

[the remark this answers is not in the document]
kilchenmann: I saw, that I added the `dsp-progress-indicator` mock to the wrong files. I'll fix it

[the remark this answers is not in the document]
kilchenmann: Fixed in 9f58652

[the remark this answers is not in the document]
tobiasschweizer: I think it really depends on when `dsp-progress-indicator` is needed. Since all async HTTP calls are mocked using `Observables` that are are actually [synchronous](https://angular.io/guide/testing-components-scenarios#testing-with-a-spy), it may not be needed.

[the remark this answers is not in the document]
tobiasschweizer: also if there are conditions in the template that load more components, it really depends on how the test looks. (interactions that change these conditions)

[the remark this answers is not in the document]
kilchenmann: I understand. May I merge it now?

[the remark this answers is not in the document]
tobiasschweizer: yes, I have already approved the PR :-)

PR #145, opened 2020-08-04
Update process of release notes

kilchenmann:
This PR splits the Gren workflow step into two variations: one runs on prerelease with the `--prerelease` option and one on release without this parameter.
The options parameter is a new feature of the latest release of [lakto/gren-action](https://github.com/lakto/gren-action/tree/v2.0.0)


kilchenmann:
For the reviewers @flavens @mdelez:
This is nothing you can test. I tested it on my [private repo](https://github.com/lakto/365).
We should merge it before next release. Then we'll see if it works the same way as it did in my test repo.

.github/workflows/main.yml
```diff
@@ -125,6 +125,17 @@ jobs:
         run: cd ${{env.dist-directory}} && npm publish --access public
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN_DASCHBOT}}
-      - uses: lakto/gren-action@v1.1.0
+      - if: "github.event.release.prerelease"
+        name: Update release notes
+        uses: lakto/gren-action@v2.0.0
```
flavens, line 130: I do not understand why we use this line and line 137 in dsp-ui

kilchenmann: You mean the gren-action package? It's a github actions package to use gren, which creates the release notes in github. It was already there before, but now in an updated version to make the difference between the prerelease and the real release. In the old version, it always overwrote the pre-release flag on Github, which is of course bad.
I forked lakto/gren-action from a action script on [Github's marketplace](https://github.com/marketplace?type=actions) to use it with the gren options

kilchenmann: We use `lakto/gren-action` in all main repos of the DaSCH (dsp-api, dsp-js-lib, dsp-ui-lib and dsp-app) by the way

PR #146, opened 2020-08-04
Add missing regex variables

flavens:
resolves [DSP-525](https://dasch.myjetbrains.com/youtrack/issue/DSP-525)


kilchenmann:
@flavens Does it make sense to add also the regexp ontology name?
`ONTOLOGYNAME_REGEX = /^(?![vV][0-9]|[0-9]|[\u00C0-\u017F]).[a-zA-Z0-9]+\S*$/;`

At the moment it's defined in ontology editor only...

flavens:
> @flavens Does it make sense to add also the regexp ontology name?
> `ONTOLOGYNAME_REGEX = /^(?![vV][0-9]|[0-9]|[\u00C0-\u017F]).[a-zA-Z0-9]+\S*$/;`
> 
> At the moment it's defined in ontology editor only...

in which repo is the file where you use ONTOLOGYNAME_REGEX?

kilchenmann:
> in which repo is the file where you use ONTOLOGYNAME_REGEX?

In dsp-app. It's part of the ontology editor ([#229](https://github.com/dasch-swiss/dsp-app/pull/229))

flavens:
> In dsp-app. It's part of the ontology editor ([#229](https://github.com/dasch-swiss/dsp-app/pull/229)). But we can keep it there. No problem

If you think that we will use it in another project or it is useful for other projects, we can move it here.

kilchenmann:
> If you think that we will use it in another project or it is useful for other projects, we can move it here.

Probably yes. It could be helpful. Because the ontology name has so many rules: https://docs.knora.org/03-apis/api-v2/knora-iris/#iris-for-ontologies-and-ontology-entities

flavens:
> Probably yes. It could be helpful. Because the ontology name has so many rules: https://docs.knora.org/03-apis/api-v2/knora-iris/#iris-for-ontologies-and-ontology-entities

done in 38052e5

PR #147, opened 2020-08-05
Delete property value

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-526

uses DSP-API v13.0.0-rc.11


mdelez:
@benjamingeer @nemmK is it possible to change DeleteValue to also contain a property member?

https://github.com/dasch-swiss/dsp-ui-lib/blob/0e97c6362fb3cc24c129a5ce21ec0387a53c3d47/projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts#L124-L152

My issue is that when adding a value to a property, I am able to filter to the exact property the value has been added to (line 129) but when a value is deleted, I only have access to the value's id and the type so I am limited to only filtering by the type (line 135) which isn't so nice because the type isn't unique and multiple properties can share the same type.

benjamingeer:
> is it possible to change DeleteValue to also contain a property member?

Sorry, I don't understand the question, I think you'll have to ask @tobiasschweizer .

> but when a value is deleted, I only have access to the value's id and the type so I am limited to only filtering by the type (line 135) which isn't so nice because the type isn't unique and multiple properties can share the same type.

I don't understand what you mean by filtering, but the ID is unique, so why not use that?

mdelez:
> > is it possible to change DeleteValue to also contain a property member?
> 
> Sorry, I don't understand the question, I think you'll have to ask @tobiasschweizer .
> 
> > but when a value is deleted, I only have access to the value's id and the type so I am limited to only filtering by the type (line 135) which isn't so nice because the type isn't unique and multiple properties can share the same type.
> 
> I don't understand what you mean by filtering, but the ID is unique, so why not use that?

It's certainly possible to loop through the entire resource and every value of each property until I find the one I'm looking for. I just don't think it looks very clean. Using the `.filter()` method on the array of resource properties allows me to quickly find what I'm looking for with one line. Basically, I have a large array and I want to filter out all of the irrelevant objects so that I'm left with a small array that contains only the property values of which the `DeleteValue.id` is located in.

For example, I'm trying to filter out everything except for what is in the red box below by filtering by the unique property id found in profDef (red arrow)
Screen Shot 2020-08-05 at 16 30 28

benjamingeer:
> It's certainly possible to loop through the entire resource....

Perhaps the JS-Lib could offer a way to get a value from a resource by ID (e.g. by maintaining its own internal index, which would map value IDs to properties), so you wouldn't have to loop through anything at all. What do you think, @tobiasschweizer ?

mdelez:
> > It's certainly possible to loop through the entire resource....
> 
> Perhaps the JS-Lib could offer a way to get a value from a resource by ID (e.g. by maintaining its own internal index, which would map value IDs to properties), so you wouldn't have to loop through anything at all. What do you think, @tobiasschweizer ?

A method such as `getPropertyValuesById` that returns the values array of a given property ID would be sweet 👍

mdelez:
@flavens UI questions:

1.  When a user clicks on a delete icon, should it immediately delete the value without any confirmation? What if the user accidentally clicks on it?
2. How should we handle a property that has no values due to a user deleting all of them? Currently if a property has no values, it is hidden completely. We can change the template to show properties that have no values and display it like so: 
Screen Shot 2020-08-06 at 10 45 03

HOWEVER, this will show a lot of properties for which we haven't made value components for:
Screen Shot 2020-08-06 at 10 47 11

I could find a way to hide these properties but it will be kind of an ugly temporary workaround

flavens:
> 1. When a user clicks on a delete icon, should it immediately delete the value without any confirmation? What if the user accidentally clicks on it?
> 2. How should we handle a property that has no values due to a user deleting all of them? Currently if a property has no values, it is hidden completely. We can change the template to show properties that have no values and display it like so:

@mdelez 
1. you have to implement a confirmation box, check the [mockup](https://drive.google.com/drive/folders/10mequzyxh9aRJu2J8gx51EHdZJcqJvE5):
Screenshot 2020-08-06 at 15 06 59

2. we should discuss it with André

kilchenmann:
> we should discuss it with André

My idea is, that we have an additional button at the top of the properties list (in the action panel) where we can toggle between "all props" and "props with values". Properties without values should look the same as properties with values. Otherwise it could be confusing when some properties looks like form elements and some not. The empty prop value can have this little "add"-icon to add a new value to the property.

mdelez:
> > we should discuss it with André
> 
> My idea is, that we have an additional button at the top of the properties list (in the action panel) where we can toggle between "all props" and "props with values". Properties without values should look the same as properties with values. Otherwise it could be confusing when some properties looks like form elements and some not. The empty prop value can have this little "add"-icon to add a new value to the property.

Should I handle this in a new PR? It doesn't exactly fit the theme of value deletion

flavens:
> Should I handle this in a new PR? It doesn't exactly fit the theme of value deletion

Yes, it could be.

kilchenmann:
> Should I handle this in a new PR? It doesn't exactly fit the theme of value deletion

I'm working on this topic already #152

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
```diff
@@ -114,6 +121,48 @@ export class DisplayEditComponent implements OnInit {
         }
     }
 
+    openDialog() {
```
flavens, line 124: Each method in TS file should have a JsDoc annotation to describe the method, e.g. resource-view component

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
```diff
@@ -114,6 +121,48 @@ export class DisplayEditComponent implements OnInit {
...
+
+        this._dspApiConnection.v2.values.deleteValue(updateRes as UpdateResource<DeleteValue>).pipe(
+        mergeMap((res: DeleteValueResponse) => {
+            // emit a ValueDeleted event to the listeners in resource-view component to trigger an update of the UI
+            this._valueOperationEventService.emit(new EmitEvent(Events.ValueDeleted, deleteVal));
+            return res.result;
+        })
+        ).subscribe(
+        () => {
+            // TODO: figure out what needs to be done here
```
flavens, line 161: more details could be helpful in this comment otherwise the subscribe() should not be here

projects/dsp-ui/src/lib/viewer/services/value-operation-event.service.ts
```diff
@@ -1,6 +1,7 @@
 import { ReadValue } from '@dasch-swiss/dsp-js';
 import { Subject, Subscription } from 'rxjs';
 import { filter, map } from 'rxjs/operators';
+import { BaseValue } from '@dasch-swiss/dsp-js/src/models/v2/resources/values/base-value';
```
flavens, line 4: wrong import path, it should only be "@dasch-swiss/dsp-js"

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
```diff
@@ -9,13 +9,15 @@ import {
...
     ApiResponseError,
+    DeleteValue,
     IHasPropertyWithPropertyDefinition,
     KnoraApiConnection,
     PropertyDefinition,
     ReadResource,
     ReadValue,
     SystemPropertyDefinition
 } from '@dasch-swiss/dsp-js';
+import { BaseValue } from '@dasch-swiss/dsp-js/src/models/v2/resources/values/base-value';
```
flavens, line 20: wrong import path (see previous comment) - it causes a warning in the terminal when you run `ng s`

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
```diff
@@ -116,15 +121,30 @@ export class ResourceViewComponent implements OnInit, OnChanges, OnDestroy {
             });
     }
 
-    updateResource(newValue?: ReadValue, isDeletion?: boolean): void {
+    updateResource(value: BaseValue, isDeletion: boolean): void {
```
flavens, line 124: a JsDoc comment should precede the method

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
mdelez: This was leftover code, however I do still need to subscribe so that the emit method of `ValueOperationEventService` is called. I've removed the unnecessary function inside the subscribe in 18faf9f

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
mdelez: added JsDocs in 18faf9f

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
mdelez: JsDoc added in 18faf9f

projects/dsp-ui/src/lib/viewer/services/value-operation-event.service.ts
mdelez: Screen Shot 2020-08-11 at 16 17 19

I think I may need to add the BaseValue to the DSP-JS exports?

flavens: yes you have to export it

mdelez: done in 68cbdf9

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
mdelez: done in 68cbdf9

mdelez: fixed in c5f29c0

PR #153, opened 2020-08-12
Error handling: submitting duplicate values

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-488

uses DSP-API v13.0.0-rc.11


mdelez:
Steps to test @flavens :
1. Add a new value to a property but give it the same value as an existing value in the property. Try to save the value and you will notice there is now an error message and the form stays open so that you can change the value to a valid value.

2. Add a new valid value. Edit this new value and give it the same value as an existing value in the property. Try to save the value and you will notice there is now an error message and the form stays open so that you can change the value to a valid value.

mdelez:
> @mdelez it looks good, the error messages are explicit.
> However, I've spotted an issue (but it is not related to this PR): for text values, when I delete a value, the view is not reloaded so the deleted value is still displayed. Have you already noticed it?

Yeah, I discovered that yesterday :) https://dasch.myjetbrains.com/youtrack/issue/DSP-550

PR #154, opened 2020-08-14
Hide system ontologies

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-544

uses DSP-API v13.0.0-rc.11


mdelez:
@kilchenmann I'm not sure if this is something that can be tested without updating the mock-ontology.ts in the js-lib to mock an ontology without resources

kilchenmann:
> @kilchenmann I'm not sure if this is something that can be tested without updating the mock-ontology.ts in the js-lib to mock an ontology without resources

I will see what we can do...

projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
```diff
@@ -124,6 +124,11 @@ export class AdvancedSearchComponent implements OnInit, OnDestroy {
     initializeOntologies(): void {
         this._dspApiConnection.v2.onto.getOntologiesMetadata().subscribe(
             (response: OntologiesMetadata) => {
+                response.ontologies = response.ontologies.filter(
+                    onto => onto.attachedToProject !== 'http://www.knora.org/ontology/knora-admin#SystemProject');
```
kilchenmann, line 128: instead of `http://www.knora.org/ontology/knora-admin#SystemProject` you can/should use the value from js-lib's Constants: `Constants.SystemProjectIRI`

mdelez: good point, fixed in cfc0583

tobiasschweizer:
@mdelez Could you walk me through this PR?

mdelez:
> @mdelez Could you walk me through this PR?

Sure thing :) Basically all this PR changes is removing the system ontologies from the drop down options in the advanced search component. I'm doing this by using .filter() and filtering out anything that has an `attachedToProject` value of `http://www.knora.org/ontology/knora-admin#SystemProject`. 

We also needed to adapt the js-lib to return a mocked ontology containing no resources (which was already created but not returned by the logic) in order to successfully write the unit tests.

projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.html
```diff
@@ -22,7 +28,7 @@
 
     <div class="select-property buttons">
         <button mat-mini-fab class="property-button add-property-button" color="primary" type="button"
-                (click)="addProperty()" [disabled]="activeOntology === undefined || activeProperties.length >= 4">
+                (click)="addProperty()" [disabled]="activeOntology === undefined || activeProperties.length >= 4 || resourceClasses?.length <= 0">
```
tobiasschweizer, line 31: What is this condition good for `|| resourceClasses?.length <= 0`?

tobiasschweizer (approved):
Ok, nice!

PR #160, opened 2020-08-25
Emit link value when clicked on

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-528

uses dsp-js v1.0.0-rc-7 and dsp-api rc-12

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -1,5 +1,5 @@
 <div *ngIf="mode === 'read'; else showForm" class="read-mode-view">
-  <p class="rm-value">{{valueFormControl.value?.label}}</p>
+  <p class="rm-value" (click)="refResClicked()">{{valueFormControl.value?.label}}</p>
   <p class="rm-comment" *ngIf="shouldShowComment">{{commentFormControl.value}}</p>
 </div>
 <ng-template #showForm>
--- a/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
@@ -25,7 +25,7 @@ import { LinkValueComponent } from './link-value.component';
 @Component({
   template: `
     <dsp-link-value #inputVal [displayValue]="displayInputVal" [mode]="mode" [parentResource]="parentResource"
-                    [propIri]="propIri"></dsp-link-value>`
+                    [propIri]="propIri" (referredResourceClicked)="refResClicked($event)"></dsp-link-value>`
 })
 class TestHostDisplayValueComponent implements OnInit {
 
@@ -35,6 +35,7 @@ class TestHostDisplayValueComponent implements OnInit {
   parentResource: ReadResource;
   propIri: string;
   mode: 'read' | 'update' | 'create' | 'search';
+  linkValueSelected: ReadLinkValue;
 
   ngOnInit() {
 
@@ -48,6 +49,10 @@ class TestHostDisplayValueComponent implements OnInit {
       this.mode = 'read';
     });
 
+  }
+
+  refResClicked(readLinkValue: ReadLinkValue) {
+    this.linkValueSelected = readLinkValue;
   }
 }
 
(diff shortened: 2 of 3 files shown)


projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
```diff
@@ -407,6 +407,15 @@ describe('LinkValueComponent', () => {
 
     }));
 
+    it('should emit the displayValue when the value is clicked on', () => {
+        spyOn(testHostComponent.inputValueComponent, 'refResClicked').and.callThrough();
```
tobiasschweizer, line 411: I only use spies on code that I mock (dependencies of the code to be tested) since I am not sure how spies change the behaviour of the code I want to test.

Instead of using a spy, you could just produce a side-effect when the `@Output` is fired:

https://github.com/dasch-swiss/dsp-ui-lib/blob/6ee18e416ffda6691185fad052db49dc8b9e7eef/projects/dsp-ui/src/lib/search/advanced-search/select-ontology/select-ontology.component.spec.ts#L14-L44

The you can just check if the emitted value has been assigned:


https://github.com/dasch-swiss/dsp-ui-lib/blob/6ee18e416ffda6691185fad052db49dc8b9e7eef/projects/dsp-ui/src/lib/search/advanced-search/select-ontology/select-ontology.component.spec.ts#L117-L133

projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.ts
```diff
@@ -65,10 +76,11 @@ export class LinkValueComponent extends BaseValueComponent implements OnInit, On
     searchByLabel(searchTerm: string) {
         // at least 3 characters are required
         if (typeof searchTerm === 'string' && searchTerm.length >= 3) {
-            this._dspApiConnection.v2.search.doSearchByLabel(searchTerm, 0, { limitToResourceClass: this.restrictToResourceClass }).subscribe(
-                (response: ReadResourceSequence) => {
-                    this.resources = response.resources;
-                });
+            this._dspApiConnection.v2.search.doSearchByLabel(
```
tobiasschweizer, line 79: reformatting?

projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
```diff
@@ -1,5 +1,5 @@
 <div *ngIf="mode === 'read'; else showForm" class="read-mode-view">
-  <p class="rm-value">{{valueFormControl.value?.label}}</p>
+  <p class="rm-value" (click)="refResClicked()">{{valueFormControl.value?.label}}</p>
```
tobiasschweizer, line 2: Could this new thing be made visible in the playground?

tobiasschweizer (changes requested):
please see my comments

projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
tobiasschweizer: maybe there could be a custom page for this where the `@Output` would be registered

projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.ts
mdelez: yeah, it was exceeding the character limit

projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.spec.ts
mdelez: fixed in 9064a60

projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
mdelez: If I'm understanding the question correctly, I don't think this would be possible since `@Output` event emissions can only be listened to by the direct parent, which is why I had to make that whole `ValueOperationEvent` service.

mdelez: Moved to a new task https://dasch.myjetbrains.com/youtrack/issue/DSP-578

tobiasschweizer (approved):
Thanks!

PR #161, opened 2020-08-26
Read mode date value era and calendar

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-568

uses uses js-lib v1.0.0-rc.9 and dsp-api v13.0.0-rc.12

### DIFF
--- a/projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
@@ -6,7 +6,9 @@ import { KnoraDate } from '@dasch-swiss/dsp-js';
 })
 export class KnoraDatePipe implements PipeTransform {
 
-    transform(date: KnoraDate, format?: string): string {
+    formattedString: string;
+
+    transform(date: KnoraDate, format?: string, precisionOptions?: 'era' | 'calendar' | 'all'): string {
         if (!(date instanceof KnoraDate)) {
             console.error('Non-KnoraDate provided. Expected a valid KnoraDate');
             return '';
@@ -14,11 +16,26 @@ export class KnoraDatePipe implements PipeTransform {
 
         switch (format) {
             case 'dd.MM.YYYY':
-                return `${this.leftPadding(date.day)}.${this.leftPadding(date.month)}.${date.year}`;
+                this.formattedString = `${this.leftPadding(date.day)}.${this.leftPadding(date.month)}.${date.year}`;
+                if (precisionOptions) {
+                    return this.addPrecision(date, this.formattedString, precisionOptions);
+                } else {
+                    return `${this.leftPadding(date.day)}.${this.leftPadding(date.month)}.${date.year}`;
+                }
             case 'dd-MM-YYYY':
-                return `${this.leftPadding(date.day)}-${this.leftPadding(date.month)}-${date.year}`;
+                this.formattedString = `${this.leftPadding(date.day)}-${this.leftPadding(date.month)}-${date.year}`;
+                if (precisionOptions) {
+                    return this.addPrecision(date, this.formattedString, precisionOptions);
+                } else {
+                    return `${this.leftPadding(date.day)}-${this.leftPadding(date.month)}-${date.year}`;
+                }
             case 'MM/dd/YYYY':
-                return `${this.leftPadding(date.month)}/${this.leftPadding(date.day)}/${date.year}`;
+                this.formattedString = `${this.leftPadding(date.month)}/${this.leftPadding(date.day)}/${date.year}`;
+                if (precisionOptions) {
+                    return this.addPrecision(date, this.formattedString, precisionOptions);
+                } else {
+                    return `${this.leftPadding(date.month)}/${this.leftPadding(date.day)}/${date.year}`;
+                }
             default:
                 return `${this.leftPadding(date.day)}.${this.leftPadding(date.month)}.${date.year}`;
         }
@@ -27,6 +44,18 @@ export class KnoraDatePipe implements PipeTransform {
     // ensures that day and month are always two digits
     leftPadding(value: number): string {
         return ('0' + value).slice(-2);
+    }
+
+    // add the era, calendar, or both to the result returned by the pipe
+    addPrecision(date: KnoraDate, value: string, precision: string): string {
+        switch (precision) {
+            case 'era':
+                return value + ' ' + date.era;
+            case 'calendar':
+                return value + ' ' + date.calendar;
+            case 'all':
+                return value + ' ' + date.era + ' ' + date.calendar;
+        }
     }
 
 }
--- a/projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
+++ b/projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
@@ -8,7 +8,7 @@ export class KnoraDatePipe implements PipeTransform {
 
     formattedString: string;
 
-    transform(date: KnoraDate, format?: string, precisionOptions?: 'era' | 'calendar' | 'all'): string {
+    transform(date: KnoraDate, format?: string, displayOptions?: 'era' | 'calendar' | 'all'): string {
         if (!(date instanceof KnoraDate)) {
             console.error('Non-KnoraDate provided. Expected a valid KnoraDate');
             return '';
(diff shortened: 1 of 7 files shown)


mdelez:
@tobiasschweizer I added the ability to show/hide labels via an input. What's not clear to me is how these input values will be chosen by the user (currently they're hard-coded).

tobiasschweizer:
> @tobiasschweizer I added the ability to show/hide labels via an input. What's not clear to me is how these input values will be chosen by the user (currently they're hard-coded).

Angular provides mechanisms that are based on the locale (https://angular.io/guide/i18n#i18n-attribute). However, I think we could just have a JSON file with the labels in different languages and use those texts in the templates instead of hardcoding them.

tobiasschweizer:
I think this is what the xlf files are for

tobiasschweizer:
you could give this a try: https://angular.io/guide/i18n#extract-the-source-language-file

mdelez:
> > @tobiasschweizer I added the ability to show/hide labels via an input. What's not clear to me is how these input values will be chosen by the user (currently they're hard-coded).
> 
> Angular provides mechanisms that are based on the locale (https://angular.io/guide/i18n#i18n-attribute). However, I think we could just have a JSON file with the labels in different languages and use those texts in the templates instead of hardcoding them.

I should be clearer 😆 I meant how will the user select the options for the inputs? Right now they are hardcoded like this
https://github.com/dasch-swiss/dsp-ui-lib/blob/e07b08e19d08057e7eb110a89763615bec25c9c0/projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts#L101-L102

which are passed to the Date Value Component `Input`'s 
https://github.com/dasch-swiss/dsp-ui-lib/blob/e07b08e19d08057e7eb110a89763615bec25c9c0/projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.html#L20

tobiasschweizer:
> > > @tobiasschweizer I added the ability to show/hide labels via an input. What's not clear to me is how these input values will be chosen by the user (currently they're hard-coded).
> > 
> > 
> > Angular provides mechanisms that are based on the locale (https://angular.io/guide/i18n#i18n-attribute). However, I think we could just have a JSON file with the labels in different languages and use those texts in the templates instead of hardcoding them.
> 
> I should be clearer 😆 I meant how will the user select the options for the inputs? Right now they are hardcoded like this
> 
> https://github.com/dasch-swiss/dsp-ui-lib/blob/e07b08e19d08057e7eb110a89763615bec25c9c0/projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts#L101-L102
> 
> which are passed to the Date Value Component `Input`'s
> 
> https://github.com/dasch-swiss/dsp-ui-lib/blob/e07b08e19d08057e7eb110a89763615bec25c9c0/projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.html#L20

the developer of an app using the lib (which I think you refer to as user) should be able to configure the component via its `@Input`s. 

My idea was that there could be an `@Input` `labels` that has a default value if the developer does not want to use it.

tobiasschweizer:
so `@Input showLabels = true` would show the labels by default, but they could also be deactivated.

projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
```diff
@@ -6,19 +6,36 @@ import { KnoraDate } from '@dasch-swiss/dsp-js';
 })
 export class KnoraDatePipe implements PipeTransform {
 
-    transform(date: KnoraDate, format?: string): string {
+    formattedString: string;
+
+    transform(date: KnoraDate, format?: string, precisionOptions?: 'era' | 'calendar' | 'all'): string {
```
tobiasschweizer, line 11: `precisionOptions` should be called differently because this is about whether to show the era and/or calendar of a given date or period, not about precision.

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
```diff
@@ -84,6 +84,10 @@ export class DisplayEditComponent implements OnInit {
     // string used as class name to add add to value-component element on hover
     backgroundColor = '';
 
+    datePrecisions: 'era' | 'calendar' | 'all';
+
+    showDateLabels = false;
```
tobiasschweizer, line 89: Do we need an initial value if this is in `ngOnInit `?

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.ts
```diff
@@ -76,7 +74,7 @@ export class DateValueComponent extends BaseValueComponent implements OnInit, On
 
         this.commentFormControl = new FormControl(null);
 
-        this.ontologyDateSetting = 'MM.dd.YYYY';
+        this.ontologyDateSetting = 'dd.MM.YYYY';
```
tobiasschweizer, line 77: Could this be defined via an `@Input` too?

tobiasschweizer: if yes, just give it `'dd.MM.YYYY'` as the default value.

projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
```diff
@@ -6,19 +6,36 @@ import { KnoraDate } from '@dasch-swiss/dsp-js';
...
         if (!(date instanceof KnoraDate)) {
             console.error('Non-KnoraDate provided. Expected a valid KnoraDate');
             return '';
         }
 
         switch (format) {
             case 'dd.MM.YYYY':
-                return `${this.leftPadding(date.day)}.${this.leftPadding(date.month)}.${date.year}`;
+                this.formattedString = `${this.leftPadding(date.day)}.${this.leftPadding(date.month)}.${date.year}`;
```
tobiasschweizer, line 19: I think when we approach [localization](https://dasch.myjetbrains.com/youtrack/issue/DSP-207), we should try to use a standard way for date formatting.

projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
```diff
@@ -29,4 +46,16 @@ export class KnoraDatePipe implements PipeTransform {
         return ('0' + value).slice(-2);
     }
 
+    // add the era, calendar, or both to the result returned by the pipe
+    addPrecision(date: KnoraDate, value: string, precision: string): string {
```
tobiasschweizer, line 50: see my comment above about naming it differently

tobiasschweizer (changes requested):

projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
mdelez: fixed in b40ef39

projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
mdelez: fixed in b40ef39

projects/dsp-ui/src/lib/action/pipes/formatting/knoradate.pipe.ts
mdelez: Yeah, this pipe has gotten a bit messier than I'd like...

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
mdelez: I wasn't exactly sure how this would be implemented in the end as I think these values should be chosen by the developer using the lib instead of hardcoded in `ngOnInit`, perhaps they could be `@Input`s?

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.ts
mdelez: I don't see why not, changed in 1004d5b

tobiasschweizer:
I have just noticed that there is a problem when the data does not have day precision (no day or no month). This means that the day and month could be `undefined`. 

A date with year precision would be displayed as "Date: ed.ed.1653" (I suppose `undefined` is turned into "ed")

mdelez:
> I have just noticed that there is a problem when the data does not have day precision (no day or no month). This means that the day and month could be `undefined`.
> 
> A date with year precision would be displayed as "Date: ed.ed.1653" (I suppose `undefined` is turned into "ed")

Interesting, and so the story continues 😅 which resource IRI are you using? or, how did you provide a date that has no day precision? Also something to note is that a user is currently not able to enter a date without a day precision because of the date picker, so I suppose this is data from a previous project (BEOL)?

mdelez:
> I have just noticed that there is a problem when the data does not have day precision (no day or no month). This means that the day and month could be `undefined`.
> 
> A date with year precision would be displayed as "Date: ed.ed.1653" (I suppose `undefined` is turned into "ed")

Also, `undefined` is turned into "ed" because of this brilliant line of code lol
```
    leftPadding(value: number): string {
        return ('0' + value).slice(-2);
    }
```

tobiasschweizer:
> > I have just noticed that there is a problem when the data does not have day precision (no day or no month). This means that the day and month could be `undefined`.
> > A date with year precision would be displayed as "Date: ed.ed.1653" (I suppose `undefined` is turned into "ed")
> 
> Interesting, and so the story continues 😅 which resource IRI are you using? or, how did you provide a date that has no day precision? Also something to note is that a user is currently not able to enter a date without a day precision because of the date picker, so I suppose this is data from a previous project (BEOL)?

I think you can just change the sample date value provided by the test resource by setting the day and/or month to `undefined` and adapt the precision (enum `Precision`) and the string representation.

You can find some examples here: https://github.com/dasch-swiss/dsp-js-lib/blob/ac25f1df7d71957004102664f1933cac5d3de3a5/src/api/v2/values/values-endpoint.spec.ts#L962-L1098

tobiasschweizer:
This is the logic the figures out the precision of a given date:
https://github.com/dasch-swiss/dsp-js-lib/blob/ac25f1df7d71957004102664f1933cac5d3de3a5/src/models/v2/resources/values/read/read-date-value.ts#L67-L76

Note that you cannot have a date with a day but no month.

mdelez:
> I have just noticed that there is a problem when the data does not have day precision (no day or no month). This means that the day and month could be `undefined`.
> 
> A date with year precision would be displayed as "Date: ed.ed.1653" (I suppose `undefined` is turned into "ed")

fixed in 08746b4. All unit tests enabled in 2d11362 😖😖😖

tobiasschweizer (approved):
Thanks for this PR! Works well in BEOL app.

PR #163, opened 2020-08-31
Anchor Tag in Link Value's HTML

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-584

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -1,6 +1,6 @@
 <div *ngIf="mode === 'read'; else showForm" class="read-mode-view">
     <p class="rm-value" (click)="refResClicked()">
-        <a>{{valueFormControl.value?.label}}</a>
+        <a class="link">{{valueFormControl.value?.label}}</a>
     </p>
   <p class="rm-comment" *ngIf="shouldShowComment">{{commentFormControl.value}}</p>
 </div>
--- a/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
+++ b/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
@@ -1,5 +1,7 @@
 <div *ngIf="mode === 'read'; else showForm" class="read-mode-view">
-  <p class="rm-value" (click)="refResClicked()">{{valueFormControl.value?.label}}</p>
+    <p class="rm-value" (click)="refResClicked()">
+        <a>{{valueFormControl.value?.label}}</a>
+    </p>
   <p class="rm-comment" *ngIf="shouldShowComment">{{commentFormControl.value}}</p>
 </div>
 <ng-template #showForm>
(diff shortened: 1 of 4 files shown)


projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
```diff
@@ -1,5 +1,7 @@
 <div *ngIf="mode === 'read'; else showForm" class="read-mode-view">
-  <p class="rm-value" (click)="refResClicked()">{{valueFormControl.value?.label}}</p>
+    <p class="rm-value" (click)="refResClicked()">
+        <a>{{valueFormControl.value?.label}}</a>
```
mdelez, line 3: Will this eventually have an `href` attribute? As it stands now, it's not clear to the user that it is something that can be clicked on. Either adding an `href` or some css to make it look like a link would help.

tobiasschweizer: I thought about this too. I theory, we could make a href using the ARK URL so it can be dereferenced. However, in the case of the BEOL app, we use the `@Output` to navigate to the resource.

About the CSS: Should the lib provide CSS or should this be left up to the app which uses the lib?

mdelez: > I thought about this too. I theory, we could make a href using the ARK URL so it can be dereferenced. However, in the case of the BEOL app, we use the `@Output` to navigate to the resource.
> 
> About the CSS: Should the lib provide CSS or should this be left up to the app which uses the lib?

I would recommend the lib provides the css. In the event that it also has an `href`, it won't look any different (assuming you make the css the same as a standard link).

tobiasschweizer: https://github.com/dasch-swiss/dsp-ui-lib/pull/163/commits/7cf63c48e580beb354542a3630ccefd54e46a79d adds the "link" class to both link and URI value. There is some CSS that defines on the lib level what links should look like, and this CSS is imported into the lib's SCSS file:

https://github.com/dasch-swiss/dsp-ui-lib/blob/46be90ad408baa0d66e425cad6a1c8bb6cf8f409/projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.scss#L1

tobiasschweizer: Anchor tags need not include the href attribute: https://stackoverflow.com/questions/10510191/valid-to-use-a-anchor-tag-without-href-attribute

However, not the `@Output` event is not bound when you use the property component. What could we do about that?

mdelez (approved):
LGTM!

projects/dsp-ui/src/lib/viewer/values/link-value/link-value.component.html
tobiasschweizer: Shall I create a DSP task for [the Output](https://github.com/dasch-swiss/dsp-ui-lib/pull/163#discussion_r480005385)?

mdelez: > Shall I create a DSP task for [the Output](https://github.com/dasch-swiss/dsp-ui-lib/pull/163#discussion_r480005385)?

yes please

tobiasschweizer: https://dasch.myjetbrains.com/youtrack/issue/DSP-586

PR #164, opened 2020-08-31
dsp-property-toolbar warnings in unit tests

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-585


mdelez:
I decided to mock the component rather than try to use the real one since there's no need to import the real property toolbar component

https://stackoverflow.com/questions/44504468/angular-2-karma-test-component-name-is-not-a-known-element

tobiasschweizer (approved):
Great, thx.!

PR #169, opened 2020-09-04
Search Panel: Advanced Search Not Displayed

flavens:
closes [DSP-590](https://dasch.myjetbrains.com/youtrack/issue/DSP-590)

### DIFF
--- a/projects/dsp-ui/src/lib/search/search-panel/search-panel.component.html
+++ b/projects/dsp-ui/src/lib/search/search-panel/search-panel.component.html
@@ -1,8 +1,8 @@
 <div class="dsp-search-panel" #fullSearchPanel cdkOverlayOrigin>
 
     <!-- DESKTOP VERSION -->
-    <dsp-fulltext-search class="dsp-fulltext-search" [projectfilter]="projectfilter"
-        [filterbyproject]="filterbyproject" (search)="search.emit($event)">
+    <dsp-fulltext-search class="dsp-fulltext-search" [projectfilter]="projectfilter" [filterbyproject]="filterbyproject"
+        (search)="search.emit($event)">
     </dsp-fulltext-search>
 
     <!-- advanced search button: if advanced === true -->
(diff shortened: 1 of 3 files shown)


projects/dsp-ui/src/lib/search/search-panel/search-panel.component.html
```diff
@@ -44,9 +44,8 @@ <h4 *ngIf="!showAdvanced">Expert search</h4>
             </span>
         </div>
         <div class="dsp-menu-content">
-            <!-- TODO: reactivate this tags when the components are ready -->
-            <!-- <dsp-advanced-search *ngIf="showAdvanced" (search)="search.emit($event); closeMenu()"></dsp-advanced-search> -->
-            <!-- <dsp-expert-search *ngIf="!showAdvanced" (search)="search.emit($event); closeMenu()"></dsp-expert-search> -->
+            <dsp-advanced-search *ngIf="showAdvanced" (search)="search.emit($event); closeMenu()"></dsp-advanced-search>
```
tobiasschweizer, line 47: Could you put `search.emit($event); closeMenu()` in a method? So then you could call in both cases: advanced and expert search.

projects/dsp-ui/src/lib/search/search-panel/search-panel.component.html
```diff
@@ -44,9 +44,8 @@ <h4 *ngIf="!showAdvanced">Expert search</h4>
...
+            <dsp-expert-search *ngIf="!showAdvanced" (search)="search.emit($event); closeMenu()"></dsp-expert-search>
```
tobiasschweizer, line 48: Shouldn't this check for the input `expert`?

tobiasschweizer (changes requested):

projects/dsp-ui/src/lib/search/search-panel/search-panel.component.html
flavens: fixed in 139009f

projects/dsp-ui/src/lib/search/search-panel/search-panel.component.html
flavens: I do not understand what you mean but the flag showAdvanced is doing the check (advanced or expert search).

tobiasschweizer: But couldn't you use [`*ngIf` with an else block](https://angular.io/api/common/NgIf#description) to make this clearer?

tobiasschweizer: I was just wondering whether the `*ngIf ... else` syntax would be clearer than two `*ngIf`s.

flavens: I would keep it this way

tobiasschweizer (approved):
ok, works for me in BEOL now.

PR #172, opened 2020-09-09
Show label for URI Value (if set)

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-588

I figured that I can use the ternary operator in the template which simplifies the logic quite a bit.

The following sources suggest that approach:
- https://angular.io/guide/template-expression-operators#the-pipe-operator- (see blue box)
- https://angular.io/api/common/NgIf#showing-an-alternative-template-using-else (see template)


mdelez (approved):
LGTM!

PR #173, opened 2020-09-10
Viewer: Use span Elements instead of div and p in Read Mode

tobiasschweizer:
resolves https://dasch.myjetbrains.com/youtrack/issue/DSP-587

This PR substitutes `<div>` and `<p>` elements for `<span>` elements in read-mode.
This should give users of the value components more flexibility when using them in an app.

For some components (date, timestamp, interval), I added some component specific CSS. I added general instructions for the `rm-value` class in the viewer's CSS since it is applied to all value components. 

However, I am still not sure how easy it would be to overwrite the styles provided by a lib on the app level.
Consider the following example from BEOL:

Screenshot 2020-09-10 at 10 28 53

Because of _viewer.scss:
```css
.read-mode-view {
  font: 400 15px/24px sans-serif;
  .rm-value {
    display: block;
  }
}
```

there is some extra space for the date value. But it is needed because otherwise the comment would be directly attached to the value.

### DIFF
--- a/projects/dsp-ui/src/assets/style/_viewer.scss
+++ b/projects/dsp-ui/src/assets/style/_viewer.scss
@@ -10,8 +10,8 @@
 
 .read-mode-view {
   font: 400 15px/24px sans-serif;
-  p {
-    margin: 0px;
+  .rm-value {
+    display: block;
   }
 }
 
--- a/projects/dsp-ui/src/assets/style/_viewer.scss
+++ b/projects/dsp-ui/src/assets/style/_viewer.scss
@@ -10,8 +10,9 @@
 
 .read-mode-view {
   font: 400 15px/24px sans-serif;
-  .rm-value {
+  .rm-value, .rm-comment {
     display: block;
+    margin: 0px
   }
 }
 
--- a/projects/dsp-ui/src/assets/style/_viewer.scss
+++ b/projects/dsp-ui/src/assets/style/_viewer.scss
@@ -12,7 +12,7 @@
   font: 400 15px/24px sans-serif;
   .rm-value, .rm-comment {
     display: block;
-    margin: 0px
+    margin: 0px;
   }
 }
 
--- a/projects/dsp-ui/src/lib/viewer/values/boolean-value/boolean-value.component.html
+++ b/projects/dsp-ui/src/lib/viewer/values/boolean-value/boolean-value.component.html
@@ -1,7 +1,7 @@
-<div *ngIf="mode === 'read'; else showForm" class="read-mode-view">
-    <p class="rm-value">{{valueFormControl.value | formattedBoolean:preferredDisplayType}}</p>
-    <p class="rm-comment" *ngIf="shouldShowComment">{{commentFormControl.value}}</p>
-</div>
+<span *ngIf="mode === 'read'; else showForm" class="read-mode-view">
+    <span class="rm-value">{{valueFormControl.value | formattedBoolean:preferredDisplayType}}</span>
+    <span class="rm-comment" *ngIf="shouldShowComment">{{commentFormControl.value}}</span>
+</span>
 <ng-template #showForm>
     <span [formGroup]="form" (ngSubmit)="onSubmit()">
         <mat-checkbox [formControlName]="'booleanValue'" class="value boolValue child-value-component">
--- a/projects/dsp-ui/src/lib/viewer/values/color-value/color-value.component.html
+++ b/projects/dsp-ui/src/lib/viewer/values/color-value/color-value.component.html
@@ -1,15 +1,15 @@
-<div *ngIf="mode === 'read'; else showForm" class="read-mode-view">
-    <p>Hex value:</p>
-    <p class="rm-value">{{ valueFormControl.value }}</p>
-    <p>Preview:</p>
+<span *ngIf="mode === 'read'; else showForm" class="read-mode-view">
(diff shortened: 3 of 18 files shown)


projects/dsp-ui/src/assets/style/_viewer.scss
```diff
@@ -10,8 +10,8 @@
 
 .read-mode-view {
   font: 400 15px/24px sans-serif;
-  p {
-    margin: 0px;
+  .rm-value {
+    display: block;
```
mdelez, line 14: I would also add `margin: 0px` inside .rm-value to align some of the more complex value component values with their label

Screen Shot 2020-09-10 at 11 01 17

mdelez (changes requested):

mdelez:
Other things I noticed:
1. I think you missed updating the color value read mode as it still uses `<p>` and looks a bit strange now
Screen Shot 2020-09-10 at 11 05 46

2. The two date-input components for a period used to be side by side but now they are stacked one on top of the other. I think the checkbox is affecting it since the interval component also uses this css but doesn't have to account for a checkbox in the middle. I would recommend moving the checkbox somewhere else in the template.

3. There is still a bit more top-padding given to the value components which prevents the values from aligning perfectly with their labels. The padding-top can be removed from the `.property-value`. (I know it wasn't you who added this)
Screen Shot 2020-09-10 at 11 16 23

https://github.com/dasch-swiss/dsp-ui-lib/blob/f1096d3c39f5115441f1a03194ca195879fcc7ab/projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.scss#L63-L66

tobiasschweizer:
> Other things I noticed:
> 
> 1. I think you missed updating the color value read mode as it still uses `<p>` and looks a bit strange now
> 
> Screen Shot 2020-09-10 at 11 05 46
> 
> 1. The two date-input components for a period used to be side by side but now they are stacked one on top of the other. I think the checkbox is affecting it since the interval component also uses this css but doesn't have to account for a checkbox in the middle. I would recommend moving the checkbox somewhere else in the template.
> 2. There is still a bit more top-padding given to the value components which prevents the values from aligning perfectly with their labels. The padding-top can be removed from the `.property-value`. (I know it wasn't you who added this)
> 
> Screen Shot 2020-09-10 at 11 16 23
> 
> https://github.com/dasch-swiss/dsp-ui-lib/blob/f1096d3c39f5115441f1a03194ca195879fcc7ab/projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.scss#L63-L66

Hm I guess I din't know what I was doing. Could you give me a hand with the CSS?

mdelez:
> Hm I guess I din't know what I was doing. Could you give me a hand with the CSS?

sure thing :) I can do it after the meeting

tobiasschweizer:
@flavie I cannot figure out why there is so much space in the list of results in BEOL (see https://github.com/dasch-swiss/dsp-ui-lib/pull/173#issue-483532403). Could we have a look at that?

mdelez:
> @Flavie I cannot figure out why there is so much space in the list of results in BEOL (see [#173 (comment)](https://github.com/dasch-swiss/dsp-ui-lib/pull/173#issue-483532403)). Could we have a look at that?
> 
> **EDIT** @mdelez is taking a look at it.

The space comes from a `<br>` in the template of `search-results.component.html`

https://github.com/dhlab-basel/beol/blob/58446345b1365c22042b04c13d07be04b0ee236e/src/app/search-results/search-results.component.html#L59-L63

This `<br>` is used to put the span with the property label on a new line since the label and value are inside of spans. I changed the value component html manually from using a `<div>` and a `<p>` to using two `<span>`s and it seems to work fine so I think these changes should fix the spacing when implemented in BEOL

Screen Shot 2020-09-11 at 14 26 57

@tobiasschweizer did you try out these changes in BEOL with yalc?

tobiasschweizer:
@mdelez I made changes to the search results component: https://github.com/dhlab-basel/beol/pull/172

Do you think that would solve the issue?

projects/dsp-ui/src/assets/style/_viewer.scss
tobiasschweizer: done in cda4f6a30af53b31cad89a9b8ff13f525d2c1da4

tobiasschweizer:
@kilchenmann @flavens I added some instructions to the viewer's CSS so the values in read mode are displayed in block style. How easy is it for an app like BEOL to change these instructions?

kilchenmann:
@tobiasschweizer What instructions? I do not understand...

tobiasschweizer:
> @tobiasschweizer What instructions? I do not understand...



> @tobiasschweizer What instructions? I do not understand...

sorry, I mean CSS https://github.com/dasch-swiss/dsp-ui-lib/blob/cda4f6a30af53b31cad89a9b8ff13f525d2c1da4/projects/dsp-ui/src/assets/style/_viewer.scss#L11-L17

tobiasschweizer:
> missing semicolon

fixed in https://github.com/dasch-swiss/dsp-ui-lib/pull/173/commits/00ce124b7a377b7aefb5f153f11211d46c41b2d3

tobiasschweizer:
@mdelez Could you please re-review this PR?

kilchenmann (approved):
It looks good to me.

mdelez (approved):
LGTM, thanks :)

PR #174, opened 2020-09-10
Confirmation Dialog Message Subcomponent

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-613

uses js-lib rc.10 and dsp-api rc.14

### DIFF
--- a/projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
+++ b/projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
@@ -105,24 +105,22 @@ describe('ConfirmationDialogComponent', () => {
         overlayContainer.ngOnDestroy();
     });
 
-    it('should display a confirmation dialog', () => {
+    it('should display a confirmation dialog', async () => {
 
         testHostComponent.openDialog();
 
         testHostFixture.detectChanges();
 
-        testHostFixture.whenStable().then(() => {
+        await testHostFixture.whenStable();
 
-            const dialogDiv = document.querySelector('mat-dialog-container');
-            expect(dialogDiv).toBeTruthy();
+        const dialogDiv = document.querySelector('mat-dialog-container');
+        expect(dialogDiv).toBeTruthy();
 
-            const dspConfirmMsg = document.querySelector('dsp-confirmation-message');
-            expect(dspConfirmMsg).toBeTruthy();
+        const dspConfirmMsg = document.querySelector('dsp-confirmation-message');
+        expect(dspConfirmMsg).toBeTruthy();
 
-            const dialogTitle = dialogDiv.querySelector('.title');
-            expect(dialogTitle.innerHTML.trim()).toEqual('Are you sure you want to delete this value from Integer?');
-
-        });
+        const dialogTitle = dialogDiv.querySelector('.title');
+        expect(dialogTitle.innerHTML.trim()).toEqual('Are you sure you want to delete this value from Integer?');
 
     });
 
--- a/projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
+++ b/projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
@@ -7,7 +7,9 @@ import { MatButtonHarness } from '@angular/material/button/testing';
 import { MAT_DIALOG_DATA, MatDialog, MatDialogModule, MatDialogRef } from '@angular/material/dialog';
 import { MatDialogHarness } from '@angular/material/dialog/testing';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { ReadIntValue } from '@dasch-swiss/dsp-js';
 import { ConfirmationDialogComponent } from './confirmation-dialog.component';
+import { ConfirmationMessageComponent } from './confirmation-message/confirmation-message.component';
 
 /**
  * Test host component to simulate parent component with a confirmation dialog.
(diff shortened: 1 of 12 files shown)


projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
```diff
@@ -50,7 +60,8 @@ describe('ConfirmationDialogComponent', () => {
         TestBed.configureTestingModule({
             declarations: [
                 ConfirmationDialogComponent,
-                ConfirmationDialogTestHostComponent
+                ConfirmationDialogTestHostComponent,
+                ConfirmationMessageComponent
```
tobiasschweizer, line 64: I wonder if it would be better to mock `ConfirmationMessageComponent ` here since this is a unit test for the confirmation dialog.

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
```diff
@@ -19,14 +21,22 @@ class ConfirmationDialogTestHostComponent {
 
     confirmationDialogResponse: string;
 
+    testValue: ReadIntValue;
+
     constructor(private dialog: MatDialog) {
     }
 
     openDialog() {
+        this.testValue = new ReadIntValue();
```
tobiasschweizer, line 30: Could you use a mocked int value from dsp-js-lib?

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
```diff
@@ -99,10 +110,19 @@ describe('ConfirmationDialogComponent', () => {
             expect(dialogDiv).toBeTruthy();
```
tobiasschweizer, line 110: Couldn't you use the harnesses also in this `it`? I think then you could get rid of the `whenStable`.

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
```diff
@@ -99,10 +110,19 @@ describe('ConfirmationDialogComponent', () => {
...
 
             const dialogTitle = dialogDiv.querySelector('.title');
-            expect(dialogTitle.innerHTML.trim()).toEqual('Title');
+            expect(dialogTitle.innerHTML.trim()).toEqual('Are you sure you want to delete this value from My label?');
+
+            const dialogMessageLabel = dialogDiv.querySelector('.message .val-label');
+            expect(dialogMessageLabel.innerHTML.trim()).toEqual('Confirming this action will delete the following value from My label:');
```
tobiasschweizer, line 116: this comes from `ConfirmationMessageComponent`, right?

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-message/confirmation-message.component.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+})
+class ConfirmationMessageTestHostComponent implements OnInit {
+
+    testValue: ReadIntValue;
+
+    constructor() {
+    }
+
+    ngOnInit() {
+        this.testValue = new ReadIntValue();
```
tobiasschweizer, line 21: Could you use the mocked int value from dsp-js-lib here?

tobiasschweizer (changes requested):

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
mdelez: Correct. Actually, I think testing for the message component contents is a bit redundant here. It should just be enough to test the message component contents in the `ConfirmationMessageComponent` spec file, right?

tobiasschweizer: Yes, I think so. I think you can just mock `ConfirmationMessageComponent` with a test component that has the same `@Input` (interface). 

If you want to be super paranoid, you could try to get a reference to the component using `hostCompDe.query(By.directive(TestConfirmationMessageComponent));` to make sure it was used correctly in `ConfirmationDialogComponent`s template (similar to what we do in the tests for `DisplayEditComponent` to make sure the correct value component was selected in the template).

tobiasschweizer:
@mdelez I think it is about time we think about starting localization ;-)

mdelez:
> @mdelez I think it is about time we think about starting localization ;-)

did my 5 new hardcoded English strings push it over the line? 😅

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
mdelez: If I use a harness here, I don't know how to select a basic `<p>` element to check if the title is displayed correctly

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-message/confirmation-message.component.spec.ts
mdelez: fixed in d9cde94

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
mdelez: fixed in d9cde94

projects/dsp-ui/src/lib/action/components/confirmation-dialog/confirmation-dialog.component.spec.ts
mdelez: I mocked the `ConfirmationMessageComponent` in d9cde94 but I'm having some issues using `hostCompDe.query(By.directive(MockConfirmationMessageComponent));` to check if it exists.

tobiasschweizer: I can try to help with that on Monday. We are having a meeting anyway.

tobiasschweizer (approved):
Ok, thanks

PR #178, opened 2020-09-15
Prerelease 1.0.0-rc.7

kilchenmann:
We have to publish new release candidate because of a mistake in my previous release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dsp-ui-lib",
-  "version": "1.0.0-rc.6",
+  "version": "1.0.0-rc.7",
   "description": "Angular library to easily create a Knora/DSP user interface",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


flavens (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dsp-ui-lib",
-  "version": "1.0.0-rc.6",
+  "version": "1.0.0-rc.7",
```
tobiasschweizer, line 3: Is this necessary (the app won't be published)?

kilchenmann: No, this is not necessary, you're right. We have always updated this version. But in the new release process this will no longer be relevant anyway. This is still the old way

tobiasschweizer:
@kilchenmann the knora-api version should have been updated, too:

https://github.com/dasch-swiss/dsp-ui-lib/blob/b8cbb9d72c4c8b600e8b24cd43841db750096da5/Makefile#L62

tobiasschweizer:
and we forgot to update the changelog

PR #179, opened 2020-09-16
docs: Update CHANGELOG

kilchenmann:
I forgot to update changelog in last release. In this PR I updated it with GREN.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,34 +1,93 @@
 # Changelog
 
+## v1.0.0-rc.7 (15/09/2020)
+
+#### Enhancements
+
+- [#172](https://github.com/dasch-swiss/dsp-ui-lib/pull/172) Show label for URI Value (if set)
+- [#168](https://github.com/dasch-swiss/dsp-ui-lib/pull/168) Message component timer
+
+#### Bug Fixes
+
+- [#171](https://github.com/dasch-swiss/dsp-ui-lib/pull/171) FIX: Delete Value UI Bug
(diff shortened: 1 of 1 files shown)


flavens (approved):
LGTM but could you leave a note to explain why we skip rc6?

kilchenmann:
> LGTM but could you leave a note to explain why we skip rc6?

Where should I add it? When we use gren to update CHANGELOG automatically, it will be overwritten. There’s anyway the question how to update CHANGELOG automatically in the future...

kilchenmann:
@flavens This could be an idea, how we can keep the CHANGELOG file up to date: In each PR we should add the current PR information to the CHANGELOG file in the corresponding section and this template: `[#PR-Nr](PR URL) PR Title`
e.g. #179 docs: Update CHANGELOG

What do you think?

flavens:
> @flavens This could be an idea, how we can keep the CHANGELOG file up to date: In each PR we should add the current PR information to the CHANGELOG file in the corresponding section and this template: `[#PR-Nr](PR URL) PR Title`
> e.g. #179 docs: Update CHANGELOG
> 
> What do you think?

yes, this is a good idea, but we will have to do it each time. I do not know if it will be the case...

kilchenmann:
> yes, this is a good idea, but we will have to do it each time. I do not know if it will be the case...

No this is not the best idea. I think the one I sent you in the private chat could be better. An automated process to update CHANGELOG file with the CI actions script. I'm still looking for a good solution. But now I will merge this PR

PR #182, opened 2020-09-17
FIX: Boolean add button

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-652

uses js-lib rc.11 and dsp-api rc.15


projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.html
```diff
@@ -36,9 +36,13 @@
...
                     <!-- Add button -->
                     <!-- temporary hack to hide add button for an XML value -->
+                    <!-- ngIf:
+                         - user has write permissions
+                         - the property does not have a cardinality of 0-1 (enum for this case is 1) or if the property does have this cardinility, ensure there isn't already a value
+                         - the property id is not the same as a property with a add-value component currently open to prevent users from continuously opening add-value forms on the same property -->
                     <div *ngIf="prop.guiDef.propertyIndex !== 'http://0.0.0.0:3333/ontology/0001/anything/v2#hasRichtext' &&
                         addButtonIsVisible &&
-                        prop.guiDef.cardinality !== 1 &&
+                        (prop.guiDef.cardinality !== 1  || prop.guiDef.cardinality === 1 && prop.values.length === 0 ) &&
```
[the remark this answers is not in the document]
tobiasschweizer: could you use the cardinality util provided by dsp-js?

if not, could we somehow add what you need to it?

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.ts
```diff
@@ -99,4 +97,24 @@ export class PropertyViewComponent implements OnInit, OnDestroy {
...
     }
+
+    /**
+     * Given a resource property, check if an add button should be displayed under the property values
+     *
+     * @param prop the resource property
+     */
+    addValueIsAllowed(prop: PropertyInfoValues): boolean {
+        // check if cardinality allows for a value to be added
+        if (prop.guiDef.cardinality !== 1 ||
```
mdelez, line 108: @tobiasschweizer How can I get an instance of a `ResourceClassDefinition` to use for `CardinalityUtil.createdValueForPropertyAllowed`? I think that's what I'd want to use instead of this logic.

tobiasschweizer: the definition should be contained in the `ReadResource`'s `entityInfo.classes`

mdelez: `entityInfo.classes` has a type of `ResourceClassDefinitionWithPropertyDefinition` which is an extension of `ResourceClassDefinition` but I'm getting an error when I try to pass it in.

tobiasschweizer: there are some methods to get only certain defs

tobiasschweizer: but I think these work only for property defs

tobiasschweizer: What kind of error do you get? Can you use type assertions to make it work?

Otherwise we can look at tit on Monday.

mdelez: Screen Shot 2020-09-18 at 15 51 33

We can look at it on Monday; here's the error message. Type assertion doesn't seem to work either and it suggests to assert it as type unknown first, which also didn't help.

tobiasschweizer: I will try this myself and get back to you later this morning.

tobiasschweizer: @mdelez Could you try this `CardinalityUtil.createValueForPropertyAllowed(prop.propDef.id, prop.values.length, this.parentResource.entityInfo.classes[this.parentResource.type]);`

mdelez: That was the missing part :) Thanks!!

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.html
```diff
@@ -36,10 +36,7 @@
                     </div>
                     <!-- Add button -->
                     <!-- temporary hack to hide add button for an XML value -->
-                    <div *ngIf="prop.guiDef.propertyIndex !== 'http://0.0.0.0:3333/ontology/0001/anything/v2#hasRichtext' &&
-                        addButtonIsVisible &&
-                        prop.guiDef.cardinality !== 1 &&
-                        propID !== prop.propDef.id">
+                    <div *ngIf="addValueIsAllowed(prop) && prop.guiDef.propertyIndex !== 'http://0.0.0.0:3333/ontology/0001/anything/v2#hasRichtext'">
```
tobiasschweizer, line 39: Why does the condition contain a property IRI from the anything test project?

tobiasschweizer: ah I get it: but then this should be more generic than `'http://0.0.0.0:3333/ontology/0001/anything/v2#hasRichtext'`, you could use the gui ele

tobiasschweizer: I think we should integrate the XML / CKeditor functionality asap, so this becomes unnecessary

mdelez: I agree, I'll make the small change to use the gui element

mdelez: fixed in 94ab6a8

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.ts
```diff
@@ -100,4 +96,30 @@ export class PropertyViewComponent implements OnInit, OnDestroy {
...
+        }
+
+        const isAllowed = CardinalityUtil.createValueForPropertyAllowed(
+            prop.propDef.id, prop.values.length, this.parentResource.entityInfo.classes[this.parentResource.type]);
+
+        // check if:
+        // cardinality allows for a value to be added
+        // value component does not already have an add value form open
+        // user has write permissions
+        if (isAllowed && this.propID !== prop.propDef.id && this.addButtonIsVisible) {
```
tobiasschweizer, line 119: What does this do? `this.propID !== prop.propDef.id`

tobiasschweizer, line 119: you can just return the whole expression, no need for an `if`

tobiasschweizer: "value component does not already have an add value form open"?

mdelez: `this.propID` is a string that is assigned a `prop.propDef.id` of the corresponding property when the add button is clicked. This logic is used to hide the add button when the add value form for that property is already open so that the user can't keep opening new add value forms. Without this logic, the user would always see the add button under the property if the cardinality allows for it.

mdelez: noob code fixed in 40cb19e

tobiasschweizer: ok, I get it now :-)

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.ts
```diff
@@ -100,4 +96,26 @@ export class PropertyViewComponent implements OnInit, OnDestroy {
...
         this.propID = undefined;
     }
+
+    /**
+     * Given a resource property, check if an add button should be displayed under the property values
+     *
+     * @param prop the resource property
+     */
+    addValueIsAllowed(prop: PropertyInfoValues): boolean {
+        // temporary until CkEditor is integrated
```
tobiasschweizer, line 106: Did we already create a DSP task for that?

tobiasschweizer (approved):

projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.ts
mdelez: I don't believe so

mdelez: I found this in my drafts. I guess I forgot to click 'create' 😅 
https://dasch.myjetbrains.com/youtrack/issue/DSP-682

PR #191, opened 2020-09-23
Color Value Polish

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-623

uses dsp-js rc.11 and dsp-api rc.15

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.scss
+++ b/projects/dsp-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.scss
@@ -1,0 +1,3 @@
+.child-input-component {
+  height: 48px;
+}
--- a/projects/dsp-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.scss
+++ b/projects/dsp-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.scss
@@ -1,6 +1,0 @@
-@import "../../../../../assets/style/viewer";
-
-::ng-deep .mat-form-field-flex,
-.color-picker-input {
-  cursor: pointer;
-}
--- a/projects/dsp-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.scss
+++ b/projects/dsp-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.scss
@@ -1,1 +1,6 @@
 @import "../../../../../assets/style/viewer";
+
+::ng-deep .mat-form-field-flex,
+.color-picker-input {
+  cursor: pointer;
+}
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,1 @@
+resolves [DSP-xy](https://dasch.myjetbrains.com/youtrack/issue/DSP-xy)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,12 +109,15 @@ jobs:
       - name: Get previous tag
         id: previoustag
         uses: "WyriHaximus/github-action-get-previous-tag@master"
+      - name: Update version
+        run: cd ${{env.dist-directory}} && npm version ${{steps.previoustag.outputs.tag}} --git-tag-version=false --commit-hooks=false
       - if: "github.event.release.prerelease"
         name: Publish pre-release to npm
-        run: cd ${{env.dist-directory}} && npm version ${{steps.previoustag.outputs.tag}} --git-tag-version=false --commit-hooks=false && npm publish --tag rc --access public
+        run: npm publish ${{env.dist-directory}} --tag rc --access public
         env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN_DASCHBOT}}
       - if: "!github.event.release.prerelease"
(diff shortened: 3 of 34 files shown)


kilchenmann:
I added new style layout for read-color-value:
Screenshot 2020-09-29 at 11 28 22

And the position of the label should also be fixed in edit and create mode:
Screenshot 2020-09-29 at 12 00 33

kilchenmann:
@mdelez What do you think? Is it done? Can someone do the review?

mdelez:
> @mdelez What do you think? Is it done? Can someone do the review?

I think the only thing I would still try to do is to center the placeholder text in the middle of the box but I know it's not very easy as I've tried as well. Other than that, it looks great! You can request a review and see what the others have to say?

kilchenmann:
> I think the only thing I would still try to do is to center the placeholder text in the middle of the box but I know it's not very easy as I've tried as well. Other than that, it looks great! You can request a review and see what the others have to say?

Yes, I also wanted to center it. But it was not possible with my previous settings. I can try again....

mdelez:
> > I think the only thing I would still try to do is to center the placeholder text in the middle of the box but I know it's not very easy as I've tried as well. Other than that, it looks great! You can request a review and see what the others have to say?
> 
> Yes, I also wanted to center it. But it was not possible with my previous settings. I can try again....

It's not a big issue so don't stress too much over it :)

kilchenmann:
It's not possible to center the label because it has also affect on the comment input field.

mdelez:
> It's not possible to center the label, as this also affects the comment input field.

No worries :) You can request a review from Flavie since she was the one who originally created this component.

projects/dsp-ui/src/lib/viewer/values/color-value/color-picker/color-picker.component.scss
```diff
@@ -1 +1,6 @@
 @import "../../../../../assets/style/viewer";
+
+::ng-deep .mat-form-field-flex,
```
flavens, line 3: instead of using `::ng-deep`, would it be possible to do it in one of the main stylesheets (src/assets/style/*)?

kilchenmann: This is what I always suggested. I can move it to ds-ui.scss...

mdelez: If you move this to dsp-ui.scss, then this will be applied to every .mat-form-field-flex in the DOM. I think I originally tried to do this in dsp-ui.scss and select it via a class name but it wasn't working.

kilchenmann: @mdelez Yes I know. But I will use the component tag name as first selector then it should work. I'm not sure...I will try

mdelez: I'm actually dealing with this same CSS scenario in this [PR](https://github.com/dasch-swiss/dsp-ui-lib/pull/198). I've found that within dsp-ui.scss, I can select custom class names as expected. However the issue arrises when I try to do the same thing in _viewer.scss. Do you know why I am not able to do this in _viewer.scss?

flavens: have you imported the _viewer file in the css file?

mdelez: Well, I feel stupid now... It works now that I've imported the _viewer file

kilchenmann: I use now the `:host` selector and it works the same way...

kilchenmann: Which version should we commit and push?

flavens: well, shouldn't we keep it consistent with the other css files? or is it something different?

kilchenmann: @mdelez Do you want to push your solution? Mine is not well done, I think...

mdelez: is the decision between `:host` and `::ng-deep` in the the color value scss file? If so, I would choose `:host`.

kilchenmann: I pushed it, but not the thing with `:host` the host is bigger than the input and the pointer was not only inside the box.
@mike You can still override my solution

mdelez: aha, what you did looks good :)

flavens (commented):
@kilchenmann why is there such a big gap between the color field and the comment field in edit and create modes?

Screenshot 2020-10-05 at 10 26 13

kilchenmann:
> @kilchenmann why is there such a big gap between the color field and the comment field in edit and create modes?
> 
> Screenshot 2020-10-05 at 10 26 13

Good question. Would this be better?
Screenshot 2020-10-05 at 10 55 41

kilchenmann:
> @kilchenmann why is there such a big gap between the color field and the comment field in edit and create modes?
> 

@flavens: resolved in 2d9b79c

flavens (approved):

kilchenmann (approved):

PR #192, opened 2020-09-23
Default Boolean Value

mdelez:
closes https://dasch.myjetbrains.com/youtrack/issue/DSP-653

uses dsp-js rc.11 and dsp-api rc.15


tobiasschweizer:
Could the `FormControl` be initialised with `false`?

tobiasschweizer:
I think it doesn't matter since it will be overwritten by `resetFormControl`.

tobiasschweizer (approved):
Ok, makes sense :-)

PR #196, opened 2020-09-24
DSP-689 Resource viewer fix style in properties toolbar

kilchenmann:
resolves [DSP-689](https://dasch.myjetbrains.com/youtrack/issue/DSP-689)


kilchenmann:
Maybe this PR is already outdated. Not sure where the problem is. It works already well in the dsp-ui playground, but we had a styling issue in dsp-app. I think I have to check this PR again....

@mdelez Wait with the review!

kilchenmann:
I don't know why we have the styling issue only in the app. I suggest to use the defined solution in this PR.
@mdelez Can you review it?

mdelez:
> I don't know why we have the styling issue only in the app. I suggest to use the defined solution in this PR.
> @mdelez Can you review it?

Something seems to be triggering the responsive css for a screen with a min-width of 768px. I'll look into it :)

Screen Shot 2020-09-25 at 16 36 05

kilchenmann:
> > I don't know why we have the styling issue only in the app. I suggest to use the defined solution in this PR.
> > @mdelez Can you review it?
> 
> Something seems to be triggering the responsive css for a screen with a min-width of 768px. I'll look into it :)
> 
> Screen Shot 2020-09-25 at 16 36 05

damn! you're right. Thank you. I will fix it in the app and close this PR without merging

PR #202, opened 2020-10-05
Handle links in XML text values

tobiasschweizer:
resolves [DSP-657](https://dasch.myjetbrains.com/youtrack/issue/DSP-657)

- [x] display external links in read-mode for standard mapping (coordinate with https://dasch.myjetbrains.com/youtrack/issue/DSP-297)
- [x] react to clicking on internal links in read-mode  for standard mapping
- [x] try to use `@HostListener`: https://github.com/dhlab-basel/beol/blob/04c11b712246236cdba732689997bb12d025c888/src/app/directives/mathjax.directive.ts#L242-L375 for internal links: target resource should be shown in the resource viewer

- add target: '_blank' to attributes for external links so that they open in a new tab  -> this might be solved in https://dasch.myjetbrains.com/youtrack/issue/DSP-297


tobiasschweizer:
@flavens @mdelez @kilchenmann 
This PR adds a directive with an `@Output` to react to click and hover events on internal links in text values with markup.

I wonder how this could be integrated in the resource viewer. Could the resource viewer propagate these events, so when used in an app, the behavior could be defined.

I think the resource viewer could have on `@Ouptput` for click events on internal links that would combine:
- clicks on an link value in read-mode
- clicks on an internal link in a text value with markup in read-mode

mdelez (approved):
Nice :)

tobiasschweizer:
@flavens I think #203 will also handle external links in HTML text values in read-mode. However, could we make sure it does not affect the markup inside CKEditor (when the HTML is being edited)?

flavens:
> @flavens I think #203 will also handle external links in HTML text values in read-mode. However, could we make sure it does not affect the markup inside CKEditor (when the HTML is being edited)?

I can check that

tobiasschweizer:
> > @flavens I think #203 will also handle external links in HTML text values in read-mode. However, could we make sure it does not affect the markup inside CKEditor (when the HTML is being edited)?
> 
> I can check that

Great, thank you. If I can help you, let me know.

PR #207, opened 2020-10-08
DSP-806 - Add the attribute 'rel' to uri and geoname value component template

flavens:
resolves [DSP-806](https://dasch.myjetbrains.com/youtrack/issue/DSP-806)


tobiasschweizer (approved):
Is this for security reasons?

https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener

flavens:
> Is this for security reasons?
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener

Yes, this is what I understood from the [tutorial](https://coryrylan.com/blog/managing-external-links-safely-in-angular) :
> To solve both of these issues any link using target="_blank" should also have a rel="noopener" attribute as well. The rel="noopener" attribute makes sure that the new browser tab does not run on the same process and prevent it from accessing window.opener. [...]
This attribute fixes our performance and security issue, but now we must manually check every link in our application to make sure if it is external that it has the appropriate attributes. Luckily for us, we can leverage Angular Directives to automate this for us.

PR #208, opened 2020-10-09
DSP-762 Error handling and notification service

kilchenmann:
resolves [DSP-762](https://dasch.myjetbrains.com/youtrack/issue/DSP-762)


projects/dsp-ui/src/lib/action/services/notification.service.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+                MatSnackBarModule
+            ],
+            providers: [
+                NotificationService
+            ]
+        });
+        service = TestBed.inject(NotificationService);
+    });
+
+    it('should be created', () => {
```
flavens, line 21: I would have added more unit tests

flavens (commented):
What about the different searches? Full text / advanced / expert

projects/dsp-ui/src/lib/action/services/notification.service.spec.ts
kilchenmann: Ah yes, this was the missing thing ...

kilchenmann:
> What about the different searches? Full text / advanced / expert

What do you mean?

flavens:
> > What about the different searches? Full text / advanced / expert
> 
> What do you mean?

well, I turn off dsp-api and tried the different kind of searches and no error notification, no message, nothing has been displayed, except the error logs in the browser console log. As a user, I would like to know what is going on, why the advanced or expert search form is not displayed or why the full text search result displays 0 result whereas. I know results exist and I have permissions to see them.

kilchenmann:
> > > What about the different searches? Full text / advanced / expert
> > 
> > 
> > What do you mean?
> 
> well, I turn off dsp-api and tried the different kind of searches and no error notification, no message, nothing has been displayed, except the error logs in the browser console log. As a user, I would like to know what is going on, why the advanced or expert search form is not displayed or why the full text search result displays 0 result whereas. I know results exist and I have permissions to see them.

I thought I already added it to advanced search. I will do that. In expert search I do not have an api connection. We can display the form. The error will happen on search which is part of search results...

projects/dsp-ui/src/lib/action/services/notification.service.ts
```diff
@@ -0,0 +1,50 @@
...
+        let duration = 4200;
+        let panelClass: string;
+
+        if (notification instanceof ApiResponseError) {
+            const status = (notification.status === 0 ? 503 : notification.status);
+
+            // get list of default http status message
+            let defaultStatusMsg = this._statusMsg.default;
+
+            message = defaultStatusMsg[status].message + ' ('+status+'): '+ defaultStatusMsg[status].description;
```
mpro7, line 32: With ES6 it is possible to use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) which gives some advantages over concatenation, including better code readability.

kilchenmann: Good point. I forgot about template literals

kilchenmann: Resolved in 9923159

projects/dsp-ui/src/lib/action/services/notification.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        if (notification instanceof ApiResponseError) {
+            const status = (notification.status === 0 ? 503 : notification.status);
+
+            // get list of default http status message
+            const defaultStatusMsg = this._statusMsg.default;
+            message = `${defaultStatusMsg[status].message} (status): ${defaultStatusMsg[status].description}`;
+
+            // display for 15s
+            duration = 15000;
```
mdelez, line 34: why do we set the duration for 15 seconds here?

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.ts
```diff
@@ -159,6 +161,10 @@ export class LoginFormComponent implements OnInit {
                 this.loginFailed = (error.status === 401 || error.status === 404);
                 this.loginErrorServer = (error.status === 0 || error.status >= 500 && error.status < 600);
 
+                if (this.loginErrorServer) {
+                    this._notification.openSnackBar(error);
+                }
+
                 this.loginSuccess.emit(false);
                 this.errorMessage = error;
```
mdelez, line 169: I think `errorMessage` can be simplified to a boolean now and renamed. The template only checks if it's defined.

mdelez (changes requested):
Nice! Works well. I just have some small remarks.

projects/dsp-ui/src/lib/action/services/notification.service.ts
```diff
@@ -0,0 +1,50 @@
...
+})
+export class NotificationService {
+
+    constructor(
+        private _snackBar: MatSnackBar,
+        private _statusMsg: StatusMsg
+    ) { }
+
+
+    // TODO: maybe we can add more parameters like:
```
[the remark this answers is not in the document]
mpro7: Why actually duration couldn't be a const?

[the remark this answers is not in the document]
kilchenmann: We can set it as a const. The question is: how long should we display a notification? 15 seconds can be to long for success messages but it's okay for the server error messages. We can use a formula as suggest in the second answer here: <https://ux.stackexchange.com/questions/11203/how-long-should-a-temporary-notification-toast-appear> 😄
@mpro7 Would you like to update this PR with best practices? Thanks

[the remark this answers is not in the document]
mpro7: According to the link above and our longest possible message (correct me if I'm wrong) "The request was a legal request, but the server is refusing to respond to it. For use when authentication is possible but has failed or not yet been provided" I suggest to set it to 5 seconds.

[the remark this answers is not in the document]
kilchenmann: Youtrack also displays a snackbar notification in case of server error; the duration there is more than 10 seconds. When we have a server error (db is not running (500) or the whole stack (503)) nothing will work anymore. In this case the notification should be there for longer time and maybe should include a link to `refresh` the page e.g. "Come back in a few minutes and try to <relaod the page>". Already discussed in DSP-710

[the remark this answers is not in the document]
mpro7: As agreed during call the duration will remain as const set to 5 seconds until next possible iteration, when the behaviour of the service will be enriched.

projects/dsp-ui/src/lib/action/components/login-form/login-form.component.ts
mpro7: Done in 7bec4cd.

projects/dsp-ui/src/lib/action/services/notification.service.ts
mpro7: Explained above and changed in 7bec4cd.

mdelez (approved):
Awesome, thanks :)

PR #210, opened 2020-10-19
DSP-881 - Value Component Base Class Changes

mdelez:
resolves [DSP-881](https://dasch.myjetbrains.com/youtrack/issue/DSP-881)


flavens (commented):
looks good, but I would also need the input parentResource as an option (parentResource?)

mdelez:
> looks good, but I would also need the input parentResource as an option (parentResource?)

added in a9edde8

mdelez:
**Important note** The original purpose of this PR was to export a class but we realized this wasn't needed but instead other changes needed to be made. Hence the discrepancy between the branch name and PR/DSP task name.

flavens (approved):

PR #214, opened 2020-10-29
DSP-759 Check if a Given Date Can Be Edited

tobiasschweizer:
resolves [DSP-759](https://dasch.myjetbrains.com/youtrack/issue/DSP-759)

Note that this PR renames `ValueTypeService` to `ValueService`.

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
@@ -3,12 +3,11 @@ import { HarnessLoader } from '@angular/cdk/testing';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { Component, Inject, Input, OnInit, ViewChild } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
-import { FormBuilder, FormControl, FormGroup, Validators } from '@angular/forms';
+import { FormBuilder, FormControl, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
 import { MatButtonHarness } from '@angular/material/button/testing';
 import { MAT_DIALOG_DATA, MatDialogModule, MatDialogRef } from '@angular/material/dialog';
 import { MatDialogHarness } from '@angular/material/dialog/testing';
 import { MatIconModule } from '@angular/material/icon';
-import { MatInputHarness } from '@angular/material/input/testing';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { By } from '@angular/platform-browser';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
@@ -29,7 +28,6 @@ import {
     ReadResource,
     ReadTextValueAsHtml,
     ReadTextValueAsString,
-    ReadTextValueAsXml,
     ReadTimeValue,
     ReadUriValue,
     ReadValue,
@@ -269,7 +267,6 @@ class TestHostDisplayValueComponent implements OnInit {
 describe('DisplayEditComponent', () => {
   let testHostComponent: TestHostDisplayValueComponent;
   let testHostFixture: ComponentFixture<TestHostDisplayValueComponent>;
-  let valueTypeService: ValueService;
 
   beforeEach(async(() => {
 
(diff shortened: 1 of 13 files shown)


tobiasschweizer:
@mdelez This PR handles precision and era, but not the calendar.

tobiasschweizer:
@mdelez sorry, too early to review: still missing the logic the prevents the editing button to show up

tobiasschweizer:
@mdelez Ok, now this PR is ready for review :-)

I decided to move the logic to value type service since it is also used there. It seemed weird to me to use logic coming from a component in a service.

projects/dsp-ui/src/lib/viewer/services/value-type.service.ts
```diff
@@ -78,6 +79,56 @@ export class ValueTypeService {
...
+     * Determines if a text can be edited using the text editor.
+     *
+     * @param textValue the text value to be checked.
+     */
+    isTextEditable(textValue: ReadTextValueAsXml): boolean {
+        return textValue.mapping === 'http://rdfh.ch/standoff/mappings/StandardMapping';
+    }
+
     /**
      * Equality checks with constants below are TEMPORARY until component is implemented.
```
tobiasschweizer, line 133: @mdelez Maybe this comment should be adapted since there are also other reasons to make a value read-only. What do you think?

mdelez: Maybe this service can also be renamed? It's becoming more general and is now more of a place to add helpful functions for the value components

mdelez (changes requested):
Nice! Just left a few comments to discuss

tobiasschweizer:
> Nice! Just left a few comments to discuss

I adapted the comment and renamed the service in https://github.com/dasch-swiss/dsp-ui-lib/pull/214/commits/b918f647d119dba0f5239dd1e7617bc1f77b932a.

I renamed the service to `ValueService` because it is not only about the value's type (the value instance is checked too).

Do you know if this service is used directly somewhere (outside the lib)? If yes, usage would break with the next release.

mdelez:
> Do you know if this service is used directly somewhere (outside the lib)? If yes, usage would break with the next release.

It is used [here](https://github.com/dasch-swiss/dsp-app/blob/4b2920d1200ea929131ba1e550f10233aba2dfb8/src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts#L36) in DSP-APP. However, the [PR](https://github.com/dasch-swiss/dsp-app/pull/324) that this code is in isn't merged yet so we can update it easily :)

tobiasschweizer:
> > Do you know if this service is used directly somewhere (outside the lib)? If yes, usage would break with the next release.
> 
> It is used [here](https://github.com/dasch-swiss/dsp-app/blob/4b2920d1200ea929131ba1e550f10233aba2dfb8/src/app/workspace/resource/resource-instance-form/select-properties/select-properties.component.ts#L36) in DSP-APP. However, the [PR](https://github.com/dasch-swiss/dsp-app/pull/324) that this code is in isn't merged yet so we can update it easily :)

Ok, great! I adapted the vars as well in https://github.com/dasch-swiss/dsp-ui-lib/pull/214/commits/338990600a5e8698425dcd2c42af5e4856646b1c

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
```diff
@@ -324,14 +322,11 @@ describe('DisplayEditComponent', () => {
         {
             provide: MatDialogRef,
             useValue: {}
-        },
-        ValueService,
```
tobiasschweizer: @mdelez I think there is no need to provide the service here since it is provided in root anyway. 
Importing `ReactiveFormsModule` fixes the injector problem with `FormBuilder`.

tobiasschweizer: However, I am not really sure what the best way is. Maybe we should spend some time on https://dasch.myjetbrains.com/youtrack/issue/DSP-98

mdelez: I'm currently writing unit tests for the PR mentioned above and the imports/injections (in test host components) for the FormBuilder are quite confusing. I think we should spend some time on what exactly needs to be imported in the unit tests that you don't get an injector error.

mdelez (approved):

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.spec.ts
tobiasschweizer: That is an excellent idea. Maybe we could try to work on https://dasch.myjetbrains.com/youtrack/issue/DSP-98 soon.

PR #217, opened 2020-10-30
Create user service

tobiasschweizer:
resolves [DSP-775](https://dasch.myjetbrains.com/youtrack/issue/DSP-775)

- [x] add user service that instantiates a `UserCache`
- [x] prevent calls for system user (standoff link created by system user)

To be done in another PR:
- [ ] add error handling in case a requests fails (https://dasch.myjetbrains.com/youtrack/issue/DSP-973, relates to https://dasch.myjetbrains.com/youtrack/issue/DSP-402)
        Currently, the failed request is stored in the cache which is not what we want.

- [ ] provide method to refresh the cache for an item (https://dasch.myjetbrains.com/youtrack/issue/DSP-972). Then this method could be used to re-request information for a specific user.


projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
```diff
@@ -128,14 +130,14 @@ export class DisplayEditComponent implements OnInit {
...
-        this._dspApiConnection.admin.usersEndpoint.getUserByIri(this.displayValue.attachedToUser).subscribe(
-            (response: ApiResponseData<UserResponse>) => {
-                this.user = response.body.user;
-            },
-            (error: ApiResponseError) => {
-                console.error(error);
-            }
-        );
+        // prevent getting info about system user (standoff link values are managed by the system)
+        if (this.displayValue.attachedToUser !== 'http://www.knora.org/ontology/knora-admin#SystemUser') {
```
tobiasschweizer, line 134: @mdelez This prevents requesting user info for the system user. Is this fine (the `user` prop) remains `undefined`.

tobiasschweizer:
@mdelez I figured that standoff links should be flagged readonly since they cannot be edited anyway. They are managed by the system.

projects/dsp-ui/src/lib/viewer/operations/display-edit/display-edit.component.ts
tobiasschweizer: The info box does not work properly for the system user:

Screenshot 2020-11-02 at 14 28 12

tobiasschweizer: fixed in https://github.com/dasch-swiss/dsp-ui-lib/pull/217/commits/089c5af762049cd44c475676127839fbf5aab9fa

mdelez (approved):
Neat :)

tobiasschweizer:
@mdelez I think it would be a good idea to mention this service someplace so that it can also be used by other components (rather than requesting users from Knora directly). What would fit best?

tobiasschweizer:
I reapplied the [optional chaining operator ](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) in https://github.com/dasch-swiss/dsp-ui-lib/pull/217/commits/98d2c34d3eac60f27997c8d957bcb341405e4b47

tobiasschweizer:
@mdelez I think the creator info (`getTooltipText`) could be generated when obtaining the user info.

When a method is called from a template like here `matTooltip="{{getTooltipText()}}"` it gets triggered many times. Instead, the method could be called when the user info is there, and its result could be saved to a class property which then could be used from the template. What do you think? Should we make a task for this?

mdelez:
> @mdelez I think the creator info (`getTooltipText`) could be generated when obtaining the user info.
> 
> When a method is called from a template like here `matTooltip="{{getTooltipText()}}"` it gets triggered many times. Instead, the method could be called when the user info is there, and its result could be saved to a class property which then could be used from the template. What do you think? Should we make a task for this?

I've noticed this as well when calling methods from the template. I see it quite a lot on StackOverflow but I'm getting the impression that it shouldn't be done? I'm open to any suggestions that prevent multiple triggers for no real reason.

tobiasschweizer:
> I've noticed this as well when calling methods from the template. I see it quite a lot on StackOverflow but I'm getting the impression that it shouldn't be done? I'm open to any suggestions that prevent multiple triggers for no real reason.

I think it is totally fine when reacting to an actual event like click etc.

In the case at hand, however, the tooltip message can be generated once and then be used in the template:

```ts

tooltipMsg = '';

onInit() {

this._userService.getUser.subscribe(
    user => this.getTooltipText()
);

}

getTooltipText(): string {
        const creationDate = 'Creation date: ' + this.displayValue.valueCreationDate;

        const creatorInfo = this.user ? '\n Value creator: ' + this.user?.givenName + ' ' + this.user?.familyName : '';

        this.tooltipMsg = creationDate + creatorInfo;
    }
```

and then in the template:

```html
<button mat-button
                        class="info"
                        matTooltip="{{getTooltipText()}}"
                        matTooltipClass="info-tooltip">
                        <mat-icon>info</mat-icon>
                </button>

```

PR #218, opened 2020-11-02
DSP-904 upload form for representations

mpro7:
resolves DSP-904

Adds upload form for StillImageRepresentation.

TODO: More tests to be added.


projects/dsp-ui/src/lib/viewer/services/upload-file.service.spec.ts
```diff
@@ -20,23 +21,54 @@ describe('UploadFileService', () => {
 
     beforeEach(() => {
         const initSpy = jasmine.createSpyObj('AppInitService', ['config']);
-        httpClientSpy = jasmine.createSpyObj('HttpClient', ['post']);
+        const httpClientSpyObj = jasmine.createSpyObj('HttpClient', ['post']);
```
tobiasschweizer, line 24: What about using [HttpTestingController](https://angular.io/api/common/http/testing/HttpTestingController)?

https://stackoverflow.com/questions/52312409/httptestingcontroller-expectone-verify-only-the-api-endpoint

tobiasschweizer: I think I got it to work in the past, it took my a while though.

mpro7: @tobiasschweizer good point. I'm trying to get this idea second go, but so far not going well.

tobiasschweizer: I think I found an example here: https://github.com/dasch-swiss/knora-ui/blob/master/projects/knora/core/src/lib/services/v2/ontology.service.spec.ts

This makes sure all HTTP requests have been handled: 

https://github.com/dasch-swiss/knora-ui/blob/2986f4e048daa28e5eb23b4920b4dea6cebebf77/projects/knora/core/src/lib/services/v2/ontology.service.spec.ts#L58-L61

For each HTTP request via the `HTTPClient`, you have to set up a thing like this:

https://github.com/dasch-swiss/knora-ui/blob/2986f4e048daa28e5eb23b4920b4dea6cebebf77/projects/knora/core/src/lib/services/v2/ontology.service.spec.ts#L82-L86

SO when you expect an HTTP request to have been sent, you check that the URL and method is correct and send back the mocked data.

tobiasschweizer (dismissed):
Sometimes the drag and drop spec fails:

>  DragDropDirective should change background-color of input on dragleave event FAILED
	Error: Expected '' to be 'rgb(242, 242, 242)'.

It seems a bit random, did you experience that too? 

Could it be related to the order in which the tests are executed? Is there some shared state between the single tests?

src/config/config.dev.json
```diff
@@ -4,5 +4,6 @@
   "apiPort": 3333,
   "apiPath": "",
   "jsonWebToken": "",
-  "logErrors": true
+  "logErrors": true,
+  "sipiUrl": "http://localhost:1024/"
```
tobiasschweizer, line 8: I think upload service will break if there is no ending slash.

mpro7:
@tobiasschweizer I think I had something similar before, but adding below spies helped:
```
spyOn(dragOver, 'preventDefault');
spyOn(dragOver, 'stopPropagation');
```
Try to log color in the test to see what you get. Also important is to do not forget about `fixture.detectChanges();` before expecting changes.

projects/dsp-ui/src/lib/viewer/directives/drag-drop.directive.ts
```diff
@@ -0,0 +1,34 @@
+import { Directive, EventEmitter, HostBinding, HostListener, Output } from '@angular/core';
+
+@Directive({
+    selector: '[dspDragDrop]'
+})
+export class DragDropDirective {
+
+    @HostBinding('style.background-color') background = '#f2f2f2';
+
+    @Output() fileDropped = new EventEmitter<any>();
```
tobiasschweizer, line 10: Could you use the actual type `FilesList` instead of `any`?

tobiasschweizer: if you want to test the directives `@Output`, this is a similar case with a test host component listening to the event:

https://github.com/dasch-swiss/dsp-ui-lib/blob/15aae792864228a85384a96b417934638cb4063b/projects/dsp-ui/src/lib/viewer/directives/text-value-html-link.directive.spec.ts#L10-L32

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
```diff
@@ -0,0 +1,134 @@
...
+        private readonly _fb: FormBuilder,
+        private readonly _ns: NotificationService,
+        private readonly _ufs: UploadFileService
+    ) { }
+
+    ngOnInit(): void {
+        this.initializeForm();
+    }
+
+    initializeForm(): void {
```
tobiasschweizer, line 35: Could you please add some documentation to the component's methods?

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
```diff
@@ -0,0 +1,134 @@
...
+import { UploadedFileResponse, UploadFileService } from '../../services/upload-file.service';
+
+@Component({
+    selector: 'dsp-upload-form',
+    templateUrl: './upload-form.component.html',
+    styleUrls: ['./upload-form.component.scss']
+})
+export class UploadFormComponent implements OnInit {
+
+    @Input() readonly resourceTyoe = 'Image'; // only StillImageRepresentation supported so far
```
tobiasschweizer, line 13: typo: `resourceTyoe`

Why is this readonly? Since it is an `@Input`, it could get a value from the parent.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
```diff
@@ -0,0 +1,134 @@
...
+    @Output() file: File;
```
tobiasschweizer, line 14: Wouldn't an `@Output` require an `EvenEmitter`?

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
```diff
@@ -0,0 +1,134 @@
...
+                this.file = null;
+            } else {
+                // show loading indicator only for files > 1MB
+                this.isLoading = this.file.size > 1048576 ? true : false;
+                formData.append(this.file.name, this.file);
+                this._ufs.upload(formData).subscribe(
+                    (res: UploadedFileResponse) => {
+                        const tempUrl = res.uploadedFiles[0].temporaryUrl;
+                        const thumbUri = `${tempUrl}/full/150,/0/default.jpg`;
+                        this.thumbnaillUrl = `${thumbUri}`.replace('http://sipi:1024/', this._ufs.envUrl);
```
tobiasschweizer, line 63: I think in production, "http://sipi:1024/" would be different.

tobiasschweizer (commented):
I tried in the playground and it worked once I was logged in :-)

Please see my comments I left.

projects/dsp-ui/src/lib/viewer/services/upload-file.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    constructor(
+        private readonly _is: AppInitService,
+        private readonly _http: HttpClient,
+        private readonly _ss: SessionService
+    ) { }
+
+    upload(file: FormData): Observable<UploadedFileResponse> {
+        const baseUrl = `${this.envUrl}upload`;
+        console.log('SESSSION', this._ss.getSession());
+        const jwt = this._ss.getSession().user.jwt;
```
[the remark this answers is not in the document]
mpro7: @tobiasschweizer good point, have never tested it while not logged in. I'm not sure how this component will be used and on which level it should be prevented. I believe whole form, which this input will be part of, will be just unavailable for not logged users? @kilchenmann cod you confirm it please?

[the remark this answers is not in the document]
tobiasschweizer: The most important thin is to perform a check if session is `null` before accessing `user`. This can then also be tested in the service's spec by passing in a session that is `null`. Let me know if can help with that.

src/config/config.dev.json
mpro7: @tobiasschweizer `sipiUrl` has ending slash, but later on in the service request the other parts of request URL are added as described [here](https://docs.knora.org/03-apis/api-v2/editing-values/#upload-files-to-sipi).

projects/dsp-ui/src/lib/viewer/directives/drag-drop.directive.ts
mpro7: Seems testing is "die unendliche Geschichte" 😃

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Do you mean comments over some lines of methods?

tobiasschweizer: the js doc style, for example:

```
    /**
     * Determines if the given value can be edited.
     *
     * @param valueTypeOrClass the type or class of the given value.
     * @param value the given value.
     */
    isReadOnly(valueTypeOrClass: string, value: ReadValue): boolean {
    ...
     }
```

in most IDEs there is some support for this with the params etc.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Thanks for that. It was readonly before it became an `@Input`. Again I'm not yet sure how this gonna work. I understand this is how the component could recognise type of resource and base on that restrict upload to the list of filetypes. I will remove it.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: I think there is no need to annotate the return type or whatever is already there in TypeScript.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: Maybe there is no need for an `@Input` as long as there is just one supported type. However, it is good that you already made a design that would support other types as well.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: This is actually good question if it depends on the env? I mean according to the docs it might, since there is **sipihost** mentioned. But base on what I got as response: `http://sipi:1024/` I assume it might be const. I will ask how does it work.

projects/dsp-ui/src/lib/viewer/services/upload-file.service.ts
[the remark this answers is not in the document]
mpro7: Sure this should be easy. I can try also to add login check before upload. Let me see.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: True, there this plugin is also available in vscode. I can install and use it, however I'm rather fan of descriptive code methodology than comments inside methods.

tobiasschweizer:
> @tobiasschweizer I think I had something similar before, but adding below spies helped:
> 
> ```
> spyOn(dragOver, 'preventDefault');
> spyOn(dragOver, 'stopPropagation');
> ```
> 
> Try to log color in the test to see what you get. Also important is to do not forget about `fixture.detectChanges();` before expecting changes.

Ok, I will spend some time on this tomorrow morning and see if I can find the reason.

mpro7:
Not sure if, but maybe adding `callThrough()` could help > `spyOn(dragLeave, 'preventDefault').and.callThrough();`

tobiasschweizer:
> Not sure if, but maybe adding `callThrough()` could help > `spyOn(dragLeave, 'preventDefault').and.callThrough();`

I will check tomorrow! Thanks for the hint.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.scss
```diff
@@ -0,0 +1,101 @@
...
+.form-container {
+  margin: 20px auto;
+  width: 50%;
+}
+.full-width {
+  width: 100%;
+}
+button {
+  cursor: pointer;
+}
```
kilchenmann, line 11: You could also use our global css class "link" on the button in the html template. But normally a button has already cursor type `pointer`.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.html
```diff
@@ -0,0 +1,63 @@
...
+                            <th>Delete</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        <tr>
+                            <td>{{ file.name }}</td>
+                            <td>{{ convertBytes(file.size) }}</td>
+                            <td>{{ convertDate(file.lastModified) }}</td>
+                            <td>
+                                <button class="delete-file"
```
kilchenmann, line 51: Why not using `mat-icon-button` or `mat-button` here?

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
```diff
@@ -0,0 +1,168 @@
...
+                    }
+                );
+            }
+
+        }
+        console.log('addFile', event, this.file, this.fileControl);
+    }
+
+    /**
+     * Converst file size to display in KM or MB
```
kilchenmann, line 86: KM? --> or KB

tobiasschweizer:
@mpro7 I fixed the method names in https://github.com/dasch-swiss/dsp-ui-lib/pull/218/commits/867e9fc180a7020cc7c25290d20dd86a9bc48bfe

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
```diff
@@ -0,0 +1,168 @@
...
+    templateUrl: './upload-form.component.html',
+    styleUrls: ['./upload-form.component.scss']
+})
+export class UploadFormComponent implements OnInit {
+
+    readonly fromLabels = {
+        upload: 'Upload file',
+        drag_drop: 'Drag and drop or click to upload'
+    };
+    readonly resourceTyoe = 'Image';
```
kilchenmann, line 17: typo in `resourceType`; but this should come from an `@Input` parameter depending on resource class. It should be of representation format defined in knora-base ontology (<https://docs.knora.org/02-knora-ontologies/knora-base/#representations>). We can use the `Constants` from js-lib here. I can help you with that...

tobiasschweizer:
I adapted the drag and drop specs in https://github.com/dasch-swiss/dsp-ui-lib/pull/218/commits/6686e49fffc3577b574353a13b999d490d9f5d9a

I add a test component to the specs that simulates the parent the directive is applied to:

https://github.com/dasch-swiss/dsp-ui-lib/blob/6686e49fffc3577b574353a13b999d490d9f5d9a/projects/dsp-ui/src/lib/viewer/directives/drag-drop.directive.spec.ts#L11-L25

Since this is a unit test for the drag and drop directive, it should not involve the upload form component:

```
 beforeEach(() => {
        fixture = TestBed.createComponent(UploadFormComponent);
```

I think the problem we saw was related to the parent because this is where the change of the background color is expected to happen. So now instead of using the actual implementation of `UploadFormComponent `, a test parent is used that applies the directive's selector.

This approach also allows for checking that the file has been emitted as expected.

src/config/config.prod.json
```diff
@@ -4,5 +4,6 @@
   "apiPort": 3333,
   "apiPath": "",
   "jsonWebToken": "",
-  "logErrors": false
+  "logErrors": false,
+  "sipiUrl": "https://iiif.dasch.swiss:1024/"
```
kilchenmann, line 8: the port is not needed here

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: > Wouldn't an `@Output` require an `EvenEmitter`?

As we spoke yesterday, `@Output` was removed with fef4c59 due to not know specification of usage.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Fixed with [this PR](https://github.com/dasch-swiss/knora-api/pull/1751) and d1699c8

projects/dsp-ui/src/lib/viewer/services/upload-file.service.ts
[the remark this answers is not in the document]
mpro7: Fixed with 679fc9b.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Fixed with 679fc9b.

src/config/config.prod.json
mpro7: According to [docs](https://docs.knora.org/03-apis/api-v2/editing-values/#upload-files-to-sipi): 
```
HTTP POST to http://sipihost/upload?token=TOKEN

The first step is to upload one or more files to Sipi, using a multipart/form-data request, where sipihost represents the host and **port** on which Sipi is running:
```

I've just put protocol, host and port in one string, but this could be separated.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.scss
mpro7: I was not aware of global styles. Cleaned a bit up those buttons with f3b60c2.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.html
mpro7: `mat-button` has something on hover which I would need to override, but `mat-icon-button` works good. Foxed with f3b60c2.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: So, @kilchenmann said we need that Input, then bringing it back with b4143c5.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Of course typo ;), fixed with b4143c5.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Brought `@Input` back and implemented simple solution based on `resourceType as string` passed from the parent with b4143c5. Have no idea which consts are you talking about, but we can either base on instance of type/class.

projects/dsp-ui/src/lib/viewer/services/upload-file.service.ts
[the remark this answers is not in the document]
tobiasschweizer: Thanks!

This was adding the `?`, right? `this._ss.getSession()?.user.jwt;`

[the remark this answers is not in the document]
mpro7: Correct

src/config/config.dev.json
tobiasschweizer: I think this is rather a convention that we should be consistent about. If the ending slash is missing, then the HTTP request would fail, right? `http://localhost:1024upload`

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: Thanks! Could you add a short explanation for the method params too (like in the example given [above](https://github.com/dasch-swiss/dsp-ui-lib/pull/218#discussion_r517890904))?

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: Ok, fine with me (without `readonly`)

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: ok

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: Thanks. So if I understand correctly, the is no hard coded expectation anymore. Instead, it comes from the config, right?

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Done with 895d3b7.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: Correct.

src/config/config.dev.json
mpro7: So as we spoke I've added some checks in 3c059f4, but was forced to edit `KnoraApiConfig` too and something is not working now. Locally all test passes and no build errors,  but here something went wrong.

tobiasschweizer: The problem is that `KnoraApiConfig` has no member `sipiUrl`. So it cannot be added to or accessed from `KnoraApiConfig`. Could you undo the commit and we will figure out what top do about this so it does not block your PR?

tobiasschweizer: and yes, it was me talking you into doing this ;-)

mpro7: I need to save this for later ;)

Anyway, just rebased the branch, so this the latest commit was removed. With this I believe we can finally close this 🙏 🚀

kilchenmann: @tobiasschweizer This is the appConfig, which should be expandable. Or is it a problem now in app-init.service?

src/config/config.prod.json
kilchenmann: It's only a detail here. The prod config will anyway be updated from server conf.

mpro7: Does it make any difference when I remove port? Should I do it so?

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.scss
kilchenmann: > I was not aware of global styles. 

Yes sure. This is something we have to document better

src/config/config.dev.json
tobiasschweizer: The problem is that `KnoraApiConfig` does not have a Sipi prop.

But it is always possible to add props to the config object and get it via `AppInitService`'s prop `config`.

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
kilchenmann: The `@Input` name should be `representation` similar to the definition from ontology and knora-api <https://docs.knora.org/02-knora-ontologies/knora-base/#representations>

Type `string` is fine.

The switch you use should go through the list of representations defined in knora-base ontology. We can access each representation definition in the `Constants` from js-lib. In this PR it would be only a case for StillImageRepresentation:
```typescript
switch (represantation) {
             case Constants.StillImageFileValue:
             ...
``` 

@tobiasschweizer is this correct? I'm not sure about the name (StillImageFileValue or HasStillImageFileValue) comes from `ResourceClass`.

tobiasschweizer: I think it depends on whether you refer to the value class or the property.

src/config/config.prod.json
kilchenmann: Yes it makes a difference, because on the defined URL this port does not exist. But as I said, this prod-conf file will anyway not be used. It's just an example

src/config/config.dev.json
kilchenmann: > The problem is that KnoraApiConfig does not have a Sipi prop.

Yes, I know. But where is it defined, that this appConfig is type of `KnoraApiConfig`? The appConfig doesn't have a type, does it?

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: Or if you refer to the resource class, it would be a `knora-api:StillImageRepresentation` or a subclass of this class.

An example would be an `anything:ThingPicture`: https://github.com/dasch-swiss/dsp-js-lib/blob/main/test/data/api/v2/values/get-still-image-file-value-response.json

src/config/config.dev.json
tobiasschweizer: The instance of `AppInitService` constructed when the app starts has two properties:

- `dspApiConfig: KnoraApiConfig`: the props needed for the `KnoraApiConfig` (only those) 

-   `config: object`: the **whole** config object (also `sipiUrl`). I would like to provide some typed access, see https://dasch.myjetbrains.com/youtrack/issue/DSP-601. Currently, you have to use this style: `_appInitService.config['sipiUrl']`.

kilchenmann: Ah yes. Thanks @tobiasschweizer for detailed explanation

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
kilchenmann: Yes `knora-api:StillImageRepresentation` this is what I thought first. Btw this constant does not exist yet in DSP-JS-Lib s. DSP-872

tobiasschweizer:
@kilchenmann I leave the final approval of this PR up to you since I mainly dealt with some questions related to testing.

I removed my [stale review](https://github.com/dasch-swiss/dsp-ui-lib/pull/218#pullrequestreview-523490305) since the issue with the test was fixed..

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
tobiasschweizer: We would have to check for subclasses too. We would use the ontology and see if a given resource class is a `knora-api:StillImageRepresentation` or a subclass.

src/config/config.dev.json
mpro7: @kilchenmann so we leave it as it is for now?

kilchenmann: yes, you can keep it as it is

projects/dsp-ui/src/lib/viewer/representation/upload-form/upload-form.component.ts
mpro7: @kilchenmann done in 199bcc6.

kilchenmann (approved):

PR #219, opened 2020-11-03
Update dependencies

kilchenmann:
- Update DSP-JS to v1.0.0-rc.15
- Update Makefile setup by using vars.mk and update README
- Prepare next release (in the previous release I have selected the wrong base branch)
- New CI feature: Update CHANGELOG on release only and attach to release assets and to build (DSP-978)


.gitignore
```diff
@@ -41,6 +41,7 @@ npm-debug.log
 yarn-error.log
 testem.log
 /typings
+CHANGELOG.md
```
tobiasschweizer, line 44: Does this mean that there is no changelog file that is part of the git repo?
So the changelog is always dynamically generated?

tobiasschweizer (commented):
Is there a way to keep the changelog in the repo and update it automatically?

.gitignore
kilchenmann: Unfortunately there is no way to keep the changelog in the repo and update it automatically. Because each change in the changelog will always need a commit and we can't commit automatically to base branch. The changelog is anyway not up-to-date in the base branch. 
We decided to generate it automatically and add it to the release assets and to the build before publishing to npm.

tobiasschweizer: So it's always accessible in the release assets as a file?

kilchenmann: Yes, exactly

tobiasschweizer: Ah, then it's totally fine.

tobiasschweizer (approved):

mdelez (approved):

PR #220, opened 2020-11-03
Propagate Click Event to Res Viewer Comp.

tobiasschweizer:
resolves DSP-784

How to test this in the playground:
- add a standoff link to the text value on the modify page:
Screenshot 2020-11-04 at 17 26 38
- from the "Viewer Module" tab you can now hover over or click standoff links (test) and links (Sierra) and get some console logs

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
+++ b/projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
@@ -14,6 +14,7 @@ import {
     IHasPropertyWithPropertyDefinition,
     KnoraApiConnection,
     PropertyDefinition,
+    ReadLinkValue,
     ReadProject,
     ReadResource,
     ReadValue,
@@ -71,6 +72,8 @@ export class ResourceViewComponent implements OnInit, OnChanges, OnDestroy {
      * @param  openProject EventEmitter which sends project information to parent component
      */
     @Output() openProject: EventEmitter<ReadProject> = new EventEmitter<ReadProject>();
+
+    @Output() referredResourceClicked: EventEmitter<ReadLinkValue> = new EventEmitter();
 
     resource: ReadResource;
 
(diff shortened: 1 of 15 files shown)


tobiasschweizer:
after discussion with @mdelez: the editing / creation of a standoff link in a text can lead to the creation of a standoff link managed by the system. We have to find a way to figure out how to get the standoff link when modifying a text value with markup

tobiasschweizer:
after discussion with @mdelez: the required logic for getting standoff link values from the resource will be implemented on a different PR so this PR can be reviewed soon.

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
```diff
@@ -72,6 +73,10 @@ export class ResourceViewComponent implements OnInit, OnChanges, OnDestroy {
      */
     @Output() openProject: EventEmitter<ReadProject> = new EventEmitter<ReadProject>();
 
+    @Output() referredResourceClicked: EventEmitter<ReadLinkValue> = new EventEmitter<ReadLinkValue>();
+
+    @Output() referredResourceHovered: EventEmitter<ReadLinkValue> = new EventEmitter<ReadLinkValue>();
```
kilchenmann, line 78: Maybe we should rename the `openProject` output to a similar name as you did e.g. `referredProjectClicked`. We should keep it consistent and I like the name `referred...Clicked`. I will create a task. We shouldn't do it in this PR

kilchenmann: Done in DSP-1023

kilchenmann (approved):

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.ts
tobiasschweizer: If you think that `referred` is a bit too verbose we could also consistently change it (being aware of breaking the public API of course).

kilchenmann: No it's fine

PR #221, opened 2020-11-04
DSP-885 Fix default gravsearch query

kilchenmann:
resolves DSP-885

The base url of an internal (and external) IRI doesn't have the leading `https` protocol. It's always `http`:
https://docs.knora.org/03-apis/api-v2/knora-iris/


kilchenmann:
@RalfBarkow Thanks for updating (1e2974f). But can you do a comment next time instead of pushing directly to a branch?

projects/dsp-ui/src/lib/search/expert-search/expert-search.component.ts
```diff
@@ -123,4 +125,16 @@ CONSTRUCT {
         return query + offsetTemplate;
     }
 
+    /**
+     * Get the IRI base url without configured api protocol.
+     * The protocol in this case is always http
+     */
+    getIriBaseUrl(): string {
+        return (
+            ('http://' + this._dspApiConfig.apiHost) +
```
RalfBarkow, line 134: Does the [Rule: quotemark](https://palantir.github.io/tslint/rules/quotemark/) apply here?

RalfBarkow: 1e2974f enforces quote character for string literals.

kilchenmann: Yes it's correct ther. But you should do a comment instead of updating the code by yourself. Thanks 😉

RalfBarkow: > do a comment instead of updating the code by yourself.

Why is that?

kilchenmann: It could 

> > do a comment instead of updating the code by yourself.
> 
> Why is that?

It could end in merge conflicts. What happens when I saw the wrong quotermarks locally too? I will change the code locally, but then I can't push because of another commit from someone else. So, I have to pull first and this could end in a merge conflict on my machine. Not everyone should push to the same branch especially not reviewers (unless it has been mutually agreed). Otherwise you have to review your own code, which doesn't make sense.

mpro7: In fact concatenation here could be replaced by ES6 template literals ;)

kilchenmann: @mpro7 You're right. I copied from js-lib. This is why I also had double quotation marks (we have other rules there). We should decide all together if we want to use ES6 literals or the "old" way. For me — and this is my opinion — the "old" way is better readable.

mpro7: @kilchenmann it is actually fine with short concatenations, but I find it really useful and more readable, especially for many properties and multi-liners. I've felt the same at the beginning as you, but since started to use template literals, I really see it more clear. [Here is few pros of using those](https://stackoverflow.com/questions/27565056/es6-template-literals-vs-concatenated-strings), because it is not only about readable but also more flexible code.

However above is just small thing, what I could find a little bit uncomfortable is no one rules set for all FE repos. But we could work on that soon.

RalfBarkow (commented):
Hoi Andrè (@kilchenmann)
Perhaps the tests (expert-search.component.spec.ts) could still be adapted?
(expected query appears several times)

kilchenmann:
> Perhaps the tests (expert-search.component.spec.ts) could still be adapted?
(expected query appears several times)

What do you mean, @RalfBarkow? Or how is it related to this issue? Can you give me an example...

RalfBarkow:
> @kilchenmann 
> What do you mean, @RalfBarkow? Or how is it related to this issue? Can you give me an example...

Sure:
dsp-ui/src/lib/search/expert-search/expert-search.component.spec.ts
* Ln 102
* Ln 135
* Ln 152
* Ln 180
* Ln 216 
* Ln 231 
sometimes with OFFSET, sometimes without.

kilchenmann:
> > @kilchenmann
> > What do you mean, @RalfBarkow? Or how is it related to this issue? Can you give me an example...
> 
> Sure:
> dsp-ui/src/lib/search/expert-search/expert-search.component.spec.ts
> 
> * Ln 102
> * Ln 135
> * Ln 152
> * Ln 180
> * Ln 216
> * Ln 231
>   sometimes with OFFSET, sometimes without.

Yes, it makes different tests. What should I do there? 
In the textarea it shouldn't have the OFFSET because this will be added automatically. And when we send the query to the API it should have an OFFSET.

RalfBarkow:
> @kilchenmann
> Yes, it makes different tests. 

That is difficult to see/read (at least for me).
Could we combine the redundant parts into one string constant?

kilchenmann:
> That is difficult to see/read (at least for me).

Each `it` is a separate step in the test &rarr; https://jasmine.github.io/api/edge/global.html#it

> Could we combine the redundant parts into one string constant?

This does not work because it needs different indentations.

Anyway: This PR is a needed bug fix, why do I have to spent now a lot of time (which we don't have) to refactor the tests?

RalfBarkow:
> @kilchenmann 
> Anyway: This PR is a needed bug fix, why do I have to spent now a lot of time (which we don't have) to refactor the tests?

You do not need to spend much time revising the tests.

RalfBarkow (approved):

projects/dsp-ui/src/lib/search/expert-search/expert-search.component.ts
```diff
@@ -35,9 +35,11 @@ export class ExpertSearchComponent implements OnInit {
     expertSearchForm: FormGroup;
     queryFormControl: FormControl;
 
+    iriBaseUrl = this.getIriBaseUrl();
```
mpro7, line 38: I think this property is redundant, since you do not use it on view, then it can be removed and in line 42 `getIriBaseUrl()` method could be directly used.

projects/dsp-ui/src/lib/search/expert-search/expert-search.component.ts
```diff
@@ -123,4 +125,16 @@ CONSTRUCT {
         return query + offsetTemplate;
     }
 
+    /**
+     * Get the IRI base url without configured api protocol.
+     * The protocol in this case is always http
+     */
+    getIriBaseUrl(): string {
```
mpro7, line 132: It is not used on view, so could be marked as `private`.

mpro7 (changes requested):

projects/dsp-ui/src/lib/search/expert-search/expert-search.component.ts
kilchenmann: Yes maybe. But I thought to move it to js-lib one day similar to the already defined `get apiUrl`. I will create a corresponding user story...

projects/dsp-ui/src/lib/search/expert-search/expert-search.component.ts
kilchenmann: Done in 014cc13

projects/dsp-ui/src/lib/search/expert-search/expert-search.component.ts
kilchenmann: Done in DSP-1036

mpro7 (approved):

PR #226, opened 2020-11-09
DSP-924 Never Import from index.ts Internally

mpro7:
resolves DSP-924

### DIFF
--- a/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
+++ b/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
@@ -22,7 +22,7 @@ import {
 } from '@dasch-swiss/dsp-js';
 import { Subscription } from 'rxjs';
 import { NotificationService } from '../../action';
-import { DspApiConnectionToken } from '../../core';
+import { DspApiConnectionToken } from '../../core/core.module';
 import { SearchParams } from '../../viewer';
 import { GravsearchGenerationService } from '../services/gravsearch-generation.service';
 import { Properties, SelectPropertyComponent } from './select-property/select-property.component';
--- a/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
+++ b/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
@@ -21,7 +21,7 @@ import {
     ResourcePropertyDefinition
 } from '@dasch-swiss/dsp-js';
 import { Subscription } from 'rxjs';
-import { NotificationService } from '../../action';
+import { NotificationService } from '../../action/services/notification.service';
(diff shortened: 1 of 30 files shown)


projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
```diff
@@ -21,9 +21,9 @@ import {
     ResourcePropertyDefinition
 } from '@dasch-swiss/dsp-js';
 import { Subscription } from 'rxjs';
-import { NotificationService } from '../../action';
-import { DspApiConnectionToken } from '../../core';
-import { SearchParams } from '../../viewer';
+import { NotificationService } from '../../action/services/notification.service';
+import { DspApiConnectionToken } from '../../core/core.module';
+import { SearchParams } from '../../viewer/';
```
tobiasschweizer, line 26: I think this should be `
import { SearchParams } from '../../viewer/views/list-view/list-view.component';`

tobiasschweizer (commented):
I think a thing that is difficult is the fact that we have one npm package (lib) consisting of several Angular modules.

> To make library code reusable you must define a public API for it. This "user layer" defines what is available to consumers of your library. A user of your library should be able to access public functionality (such as NgModules, service providers and general utility functions) through a single import path.

> The public API for your library is maintained in the public-api.ts file in your library folder. Anything exported from this file is made public when your library is imported into an application. Use an NgModule to expose services and components

https://angular.io/guide/creating-libraries#getting-started

And since we have only on npm package, I think that also inside the lib we should import from the file (and not from the `index.ts` representing the module).

tobiasschweizer:
Maybe we could think about making a sub-entry for each module, https://medium.com/@tomastrajan/the-best-way-to-architect-your-angular-libraries-87959301d3d3. However, I would first have to understand this in general.

projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
mpro7: Thanks, somehow missed it. Done in c7971bd.

mpro7:
> Maybe we could think about making a sub-entry for each module, https://medium.com/@tomastrajan/the-best-way-to-architect-your-angular-libraries-87959301d3d3. However, I would first have to understand this in general.

As you wrote, and as we spoke. This could be a second step.

tobiasschweizer:
> As you wrote, and as we spoke. This could be a second step.

Indeed :-)

I am reading this now: https://medium.com/angular-in-depth/improve-spa-performance-by-splitting-your-angular-libraries-in-multiple-chunks-8c68103692d0

tobiasschweizer (approved):
Ok, looks good!

Let's learn about splitting up the lib in different sub-entries per module.

PR #232, opened 2020-11-19
Add missing decorator to ValueOperationEventService

tobiasschweizer:
This PRs adds the missing decorator to `ValueOperationEventService`.

https://angular.io/api/core/Injectable#usage-notes:

> Marking a class with @Injectable ensures that the compiler will generate the necessary metadata to create the class's dependencies when the class is injected.


projects/dsp-ui/src/lib/viewer/services/value-operation-event.service.ts
```diff
@@ -9,6 +10,7 @@ import { filter, map } from 'rxjs/operators';
  * The ValueOperationEventService essentially creates a direct communication channel between
  * the emitting component and the listening component.
  */
+@Injectable() // must be provided on component level, i.d. resource view component.
```
mdelez, line 13: i.d. -> i.e. ?

mdelez (changes requested):

mdelez (approved):

PR #233, opened 2020-11-19
Determine from ontology if a property is read-only

tobiasschweizer:
resolves DSP-974 and DSP-1043

`ValueService.isReadOnly` has an additional third argument now: `propertyDef: ResourcePropertyDefinition`.
I added the flag "breaking" because this method could be used directly by the app.

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
@@ -317,6 +317,37 @@ describe('PropertyViewComponent', () => {
             expect(propertyViewComponentDe.query(By.css('.add-value'))).toBeDefined();
 
         });
+
+        it('should determine that adding a standoff link value is not allowed', () => {
+
+            const standoffLinkVal = testHostComponent.propArray.filter(
+                propVal => propVal.propDef.id === 'http://api.knora.org/ontology/knora-api/v2#hasStandoffLinkToValue'
+            );
+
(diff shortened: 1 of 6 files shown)


projects/dsp-ui/src/lib/viewer/views/property-view/property-view.component.spec.ts
```diff
@@ -281,17 +281,21 @@ describe('PropertyViewComponent', () => {
...
 
             testHostFixture.detectChanges();
 
-            // now the boolean property should have an add button so the amount of add buttons on the page should increase by 1
+            // now the boolean property should have an add button
+            // so the amount of add buttons on the page should increase by 1
+            // standoff links value and has incoming link value are system props and cannot be added: -2
             addButtons = propertyViewComponentDe.queryAll(By.css('button.create'));
-            expect(addButtons.length).toEqual(20);
+            expect(addButtons.length).toEqual(18);
```
mdelez, line 298: Is there a better way to know the number of expected buttons than to write a hardcoded number? Maybe we can use the propArray to check how many properties are not read-only and use that? This way if the test data changes, we don't have to keep manually updating the number. Idk how it will work with cardinalities though.

tobiasschweizer: I added some explicit tests in https://github.com/dasch-swiss/dsp-ui-lib/pull/233/commits/186d0246daea55f5c9b547de4eac5aa5a1346062

flavens (approved):

PR #234, opened 2020-11-19
DSP-1097 - Bump JS-Lib Version and Fix Breaking Changes

mdelez:
resolves DSP-1097

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1597,9 +1597,9 @@
       }
     },
     "@dasch-swiss/dsp-js": {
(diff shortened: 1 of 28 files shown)


tobiasschweizer (approved):
Looks good!

The change in package-lock.json seems ok because `npm i` took the latest available (patch) version of `canonicalize ` (defined by `~` or `^` in some dep I suppose, see https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json)

PR #237, opened 2020-12-01
DSP-1110 Bug fixes and styling improvement in resource viewer

kilchenmann:
resolves DSP-1112, DSP-1023 and part of DSP-1110

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
+++ b/projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
@@ -2,12 +2,11 @@
 
     <!-- dsp-resource-representation -->
     <div class="center">
-        <dsp-still-image *ngIf="stillImageRepresentations" [images]="stillImageRepresentations">
+        <dsp-still-image *ngIf="stillImageRepresentations.length" [images]="stillImageRepresentations">
         </dsp-still-image>
         <!-- TODO: here we'll add more viewers and players dsp-moving-image, dsp-audio etc. -->
         <mat-divider></mat-divider>
     </div>
-
 
     <!-- dsp-property-toolbar with action buttons and project information -->
     <dsp-property-toolbar *ngIf="showToolbar"
--- a/projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
+++ b/projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
@@ -10,7 +10,12 @@
 
 
     <!-- dsp-property-toolbar with action buttons and project information -->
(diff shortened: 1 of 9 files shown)


flavens (approved):
it looks good, even though

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
```diff
@@ -2,15 +2,19 @@
 
     <!-- dsp-resource-representation -->
     <div class="center">
-        <dsp-still-image *ngIf="stillImageRepresentations" [images]="stillImageRepresentations">
+        <dsp-still-image *ngIf="stillImageRepresentations.length" [images]="stillImageRepresentations">
```
mdelez, line 5: could this ever lead to a "Cannot read property 'length' of undefined" error?

mdelez (approved):
Nice :) Just a small question

projects/dsp-ui/src/lib/viewer/views/resource-view/resource-view.component.html
kilchenmann: I reset the stillImageRepresentation always as empty array before getting the data. So it shouldn't end in the mentioned error. But we have to improve this representation-thing anyway. As soon we'll have other representation it could be messy. At the moment I do not have better solution.

tobiasschweizer: Is this related to change detection? In order to trigger change detection for a complex input, a new object / array has to be assigned.

tobiasschweizer: have you tried this:

https://github.com/dasch-swiss/dsp-ui-lib/blob/ce738efb2dcf65f9ecd2cbf583241cd49968cd94/projects/dsp-ui/src/lib/viewer/representation/still-image/still-image.component.ts#L157-L167

kilchenmann: It's in case you switch between a resource with fileRepresentation and a resource without. In one case we have to display `<dsp-still-image>` and in the other we don't have to.

tobiasschweizer: > I reset the stillImageRepresentation always as empty array before getting the data.

So I guess this is correct. I assume this is done in `ngOnChanges` when the resource Iri Input changes, right?

kilchenmann: exactly

PR #238, opened 2020-12-02
DSP-1132 Advanced search limit to project

kilchenmann:
resolves DSP-1132

In this PR I added `limitToProject` parameter to advanced search. It can be used in project specific apps to list project specific ontologies only.

`limitToProject` can be used in advanced-search, in fulltext-search and in search-panel which is a combination of advanced- and fulltext-search. ~~But at the moment I have an issue with communication between the three components.~~

~~In case of `projectFilter = true` it shows a dropdown menu as part of fulltext-search to select a project. When I select a project it emits the data from fulltext-search to parent, in this case `search-panel`. limitToProject is a two-way-binding object of fulltext-search. But it ends in an `ExpressionChangedAfterItHasBeenCheckedError` in case a project is already selected when page refreshs (onInit).~~

Screen Shot 2020-12-02 at 15 28 15
**Issue solved in 9e62026**

Additional to this fix I set `filterbyproject` parameter in fulltext-search and in search-panel as deprecated. It's replaced by `limitToProject` parameter to keep it consistent. `filterbyproject` was confusing anyway, besides `projectfilter` parameter.

You can play around with the parameters in the search playground: <http://localhost:4200/search>

### DIFF
--- a/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
+++ b/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
@@ -3,7 +3,6 @@ import {
     EventEmitter,
     Inject,
     Input,
-    OnChanges,
     OnDestroy,
     OnInit,
     Output,
--- a/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
+++ b/projects/dsp-ui/src/lib/search/advanced-search/advanced-search.component.ts
@@ -2,6 +2,8 @@ import {
     Component,
     EventEmitter,
     Inject,
+    Input,
+    OnChanges,
     OnDestroy,
     OnInit,
     Output,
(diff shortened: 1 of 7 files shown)


kilchenmann:
Does someone have an idea to resolve the `ExpressionChangedAfterItHasBeenCheckedError` error I have? @flavens @mpro7 @mdelez

tobiasschweizer:
> Does someone have an idea to resolve the `ExpressionChangedAfterItHasBeenCheckedError` error I have? @flavens @mpro7 @mdelez

it means that in your parent template an expression was `undefined` and then probably after the selection of the project in some child component it is set to a project IRI. The error occurs when a child changes the state of the parent template, see https://indepth.dev/posts/1001/everything-you-need-to-know-about-the-expressionchangedafterithasbeencheckederror-error

The solution is to do the change in a resolved promise in the child component, so then it is executed as an async task.

tobiasschweizer:
I had this problem in the advanced search because the parent form is passed down to child components which then attach to it, influencing the form's validity.

So for example, if a child component adds a control which is required but has no init value, then the whole form becomes invalid. If it was valid before and the validity prop was used in the parent's template, then you get the `ExpressionChangedAfterItHasBeenCheckedError`.

kilchenmann:
@tobiasschweizer Thanks for your help. It was really helpful

tobiasschweizer:
> @tobiasschweizer Thanks for your help. It was really helpful

glad to hear

kilchenmann:
For the reviewers: Please have also a look at the following comment and the attached screen records 😉 
<https://dasch.myjetbrains.com/youtrack/issue/DSP-1132#focus=Comments-4-2178.0-0>

mdelez (approved):
LGTM!

PR #242, opened 2020-12-07
Update dsp-js-lib

kilchenmann:
Update to latest rc of DSP-JS-Lib

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1597,9 +1597,9 @@
       }
     },
     "@dasch-swiss/dsp-js": {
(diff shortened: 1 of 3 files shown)


tobiasschweizer (approved):
Looks good.

DSP-API version is still the same, right?

mpro7 (approved):
👍

kilchenmann:
> Looks good.
> 
> DSP-API version is still the same, right?

It's still the same. rc.24 of DSP-Api will be the real release and we have to go with this version

PR #250, opened 2021-01-04
Fix Value Form Control Name

tobiasschweizer:
resolves DSP-1186


tobiasschweizer:
@mdelez We missed this one in https://github.com/dasch-swiss/dsp-ui-lib/pull/250.

mdelez (approved):
Nice find :) searching for "this.form.controls" returns some results but I think they're okay

tobiasschweizer:
> Nice find :) searching for "this.form.controls" returns some results but I think they're okay

I got some results for `StringLiteralInputComponent` and `SearchListValueComponent`, and these were not affected by #250

PR #251, opened 2021-01-07
fix(Date Value Validators): Propagate valueRequiredValidator to child component (DSP-1188)

mdelez:
resolves DSP-1188

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
@@ -648,7 +648,12 @@ describe('DateValueComponent', () => {
     });
 
     it('should not create an empty value', () => {
-        expect(testHostComponent.inputValueComponent.getNewValue()).toEqual(false);
+        expect(testHostComponent.inputValueComponent.getNewValue()).toBe(false);
+        expect(testHostComponent.inputValueComponent.form.valid).toBe(true);
+    });
+
+    it('should propagate valueRequiredValidator to child component', () => {
+        expect(testHostComponent.inputValueComponent.valueRequiredValidator).toBe(false);
     });
 
   });
--- a/projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
@@ -36,6 +36,7 @@ class TestDateInputComponent implements ControlValueAccessor, MatFormFieldContro
   @Input() required: boolean;
   @Input() shouldLabelFloat: boolean;
   @Input() errorStateMatcher: ErrorStateMatcher;
+  @Input() valueRequiredValidator = true;
 
   errorState = false;
   focused = false;
@@ -116,6 +117,26 @@ class TestHostCreateValueComponent implements OnInit {
   }
 }
 
+/**
+ * Test host component to simulate parent component.
+ */
+@Component({
+    template: `
+      <dsp-date-value #inputVal [mode]="mode" [valueRequiredValidator]="false"></dsp-date-value>`
+  })
+  class TestHostCreateValueNoValueRequiredComponent implements OnInit {
+
+    @ViewChild('inputVal') inputValueComponent: DateValueComponent;
+
+    mode: 'read' | 'update' | 'create' | 'search';
+
+    ngOnInit() {
+
+      this.mode = 'create';
+
+    }
+  }
+
 describe('DateValueComponent', () => {
 
   beforeEach(async(() => {
@@ -130,6 +151,7 @@ describe('DateValueComponent', () => {
         TestDateInputComponent,
         TestHostDisplayValueComponent,
         TestHostCreateValueComponent,
+        TestHostCreateValueNoValueRequiredComponent,
         KnoraDatePipe
       ]
     })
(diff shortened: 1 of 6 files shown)


mdelez:
@tobiasschweizer efdd557 Fixes the issue for the date value component. We know we will have this issue for other complex value components as well. Should I fix them in this PR or open a new PR for each component?

P.S. I made a small fix to `int-value-component.spec` in this commit as well 🙊

flavens:
> @tobiasschweizer @flavens [efdd557](https://github.com/dasch-swiss/dsp-ui-lib/commit/efdd55731160c057a7adee0ff5fb3435552f578f) Fixes the issue for the date value component. We know we will have this issue for other complex value components as well. Should I fix them in this PR or open a new PR for each component?

It would be easier to do it in one PR. @tobiasschweizer ?

tobiasschweizer:
One PR seems less complicated.

mdelez:
After talking with @tobiasschweizer about the complexity of all of this, we have decided to split it into one PR per value component to make the reviewing process easier.

tobiasschweizer:
@mdelez I'll do the review one Monday

mdelez:
For the reviewers, to test: 
 - add `[valueRequiredValidator]="false"` to `dsp-date-value` in `add-value.component.html`
 - check that when you click on the add button in the playground for a new date value, you shouldn't see any error messages initially and the save icon should be enabled (although it will fail on knora's end if you click it)
 - change `[valueRequiredValidator]="false"` to true
 - check that the error messages behave correctly and that the save icon is disabled

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
```diff
@@ -116,6 +117,26 @@ class TestHostCreateValueComponent implements OnInit {
   }
 }
 
+/**
```
tobiasschweizer, line 120: Could you also add a test to `projects/dsp-ui/src/lib/viewer/values/date-value/date-input/date-input.component.spec.ts`? Maybe just check for the form to be valid.

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
```diff
@@ -611,5 +633,24 @@ describe('DateValueComponent', () => {
...
+      testHostFixture = TestBed.createComponent(TestHostCreateValueNoValueRequiredComponent);
+      testHostComponent = testHostFixture.componentInstance;
+      testHostFixture.detectChanges();
+
+      expect(testHostComponent).toBeTruthy();
+      expect(testHostComponent.inputValueComponent).toBeTruthy();
+    });
+
+    it('should not create an empty value', () => {
+        expect(testHostComponent.inputValueComponent.getNewValue()).toEqual(false);
```
tobiasschweizer, line 651: You could also a test to check that `valueRequiredValidator` is correctly propagated to `TestDateInputComponent` (binding in `projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.html`).

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
```diff
@@ -611,5 +633,24 @@ describe('DateValueComponent', () => {
...
+    beforeEach(() => {
+      testHostFixture = TestBed.createComponent(TestHostCreateValueNoValueRequiredComponent);
+      testHostComponent = testHostFixture.componentInstance;
+      testHostFixture.detectChanges();
+
+      expect(testHostComponent).toBeTruthy();
+      expect(testHostComponent.inputValueComponent).toBeTruthy();
+    });
+
+    it('should not create an empty value', () => {
```
tobiasschweizer, line 650: Maybe also check that the form is valid.

tobiasschweizer (commented):
See my comments on testing.

flavens (commented):
I have a question: when you check the box "is a time period", we get an error message to indicate how to fill in the fields. I guess I will get the same problem in my PR if a user checks this box... Is there a possibility to turn it into an hint instead of an error??

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
mdelez: done in c05e0fd

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
mdelez: doen in c05e0fd

projects/dsp-ui/src/lib/viewer/values/date-value/date-value.component.spec.ts
mdelez: done in c05e0fd

mdelez:
> I have a question: when you check the box "is a time period", we get an error message to indicate how to fill in the fields. I guess I will get the same problem in my PR if a user checks this box... Is there a possibility to turn it into an hint instead of an error??

I fixed the logic in d113e0b. Thanks for pointing it out :) I kept them as error messages but now they only appear when they should.

mdelez:
> maybe you could add the same test case as above in "'should mark the form's validity correctly'", however this time it is valid.

done in 2f243dd

tobiasschweizer (approved):
Looks good!

PR #253, opened 2021-01-08
fix(Interval Value Validators): Propagate valueRequiredValidator to child component (DSP-1193)

mdelez:
resolves DSP-1193


mdelez:
@tobiasschweizer same thing as #251 but for the interval value component

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.html
```diff
@@ -11,4 +11,10 @@
       End is <strong>required</strong>.
     </mat-error>
   </mat-form-field>
+
+  <div class="date-form-error">
+    <mat-error *ngIf="((startIntervalControl.hasError('startEndSameTypeRequired') || endIntervalControl.hasError('startEndSameTypeRequired')))">
```
tobiasschweizer, line 16: Is there a need for null checks for start or end as in case of the date?
Or is this taken care of by `startEndSameTypeValidator`?

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.ts
```diff
@@ -27,6 +27,20 @@ export class IntervalInputErrorStateMatcher implements ErrorStateMatcher {
     }
 }
 
+/** Interval must have a start and end of the same type, either both numbers or both null */
+export function startEndSameTypeValidator(otherInterval: FormControl): ValidatorFn {
+    return (control: AbstractControl): { [key: string]: any } | null => {
+
+        let invalid = true;
+        if (typeof(control.value) === 'number' && typeof(otherInterval.value) === 'number' ||
```
tobiasschweizer, line 35: Couldn't you just check for `null`?

Something like `invalid = !(control.value === null && otherInterval.value === null || control.value !== null && otherInterval.value !== null)` 

 I think the `input` only allows for numeric values.

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.ts
```diff
@@ -157,6 +183,16 @@ export class IntervalInputComponent extends _MatInputMixinBase implements Contro
         }
     }
 
+    ngOnInit() {
+        if (this.valueRequiredValidator) {
+            this.startIntervalControl.setValidators([Validators.required, startEndSameTypeValidator(this.endIntervalControl)]);
```
tobiasschweizer, line 188: Would it make sense to add the `startEndSameTypeValidator` also when the required validator is not set?

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.spec.ts
```diff
@@ -121,4 +150,45 @@ describe('InvertalInputComponent', () => {
...
+        expect(testHostComponent.intervalInputComponent.valueRequiredValidator).toBe(true);
+        expect(testHostComponent.intervalInputComponent.form.valid).toBe(true);
+
+        testHostComponent.intervalInputComponent.startIntervalControl.setValue(null);
+
+        testHostComponent.intervalInputComponent._handleInput();
+
+        expect(testHostComponent.intervalInputComponent.form.valid).toBe(false);
+    });
+
```
tobiasschweizer, line 163: Could you add a test case for the `startEndSameTypeValidator`? Or is this not possible because the required operator will always trigger anyway (start or end is `null`)?

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.html
mdelez: this is handled in `startEndSameTypeValidator`

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.ts
mdelez: Definitely, I forgot. Added in 3c0a216

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.ts
mdelez: This was me being overly paranoid. Refactored in 3c0a216

projects/dsp-ui/src/lib/viewer/values/interval-value/interval-input/interval-input.component.spec.ts
mdelez: done in 19f996d

mdelez:
@flavens With these changes, you can make an interval **not** required. However you will still see an error message if the interval is not entered correctly (i.e. only the start interval has been entered). There isn't really an intuitive way to disregard this error if the interval is not required because we want to ensure the user enters a valid interval. If the user does not want to enter an interval (both inputs are empty/null), that is allowed and the form will submit.

Screen Shot 2021-01-13 at 10 26 16

In this screenshot, the required validator is set to "false" however the interval only has a start so the user see's an error message. No interval will be submitted to Knora in this state.

As for the Create Resource Form, I would recommend to scroll the user to this component if the user has filled it in only partially even if it is not required. The user should fill in the form correctly by either providing a start and an end or by leaving both inputs empty. The incorrect input will have the `ng-invalid` class.

tobiasschweizer (approved):

flavens (approved):

PR #264, opened 2021-01-21
feat(viewer): integrate file upload component in form logic (DSP-1257)

tobiasschweizer:
resolves DSP-1257

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/representation/upload-file/upload-file.component.ts
+++ b/projects/dsp-ui/src/lib/viewer/representation/upload-file/upload-file.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit } from '@angular/core';
+import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { FormBuilder, FormControl, FormGroup, Validators } from '@angular/forms';
 import { Constants } from '@dasch-swiss/dsp-js';
 import { NotificationService } from '../../../action/services/notification.service';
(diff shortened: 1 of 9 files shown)


projects/dsp-ui/src/lib/viewer/representation/upload-file/upload-file.component.ts
```diff
@@ -100,29 +110,45 @@ export class UploadFileComponent implements OnInit {
...
+                }
+            }
+        );
+
         this.form = this._fb.group({
-            file: [null, Validators.required]
+            file: this.fileControl
         }, { updateOn: 'blur' });
+
+        if (this.parentForm !== undefined) {
```
mpro7, line 147: Is this in purpose just for undefined? Otherwise it can be simplified to `if (this.parentForm) {}`

mpro7 (approved):
Good job! Thanks.

projects/dsp-ui/src/lib/viewer/representation/upload-file/upload-file.component.ts
tobiasschweizer: Of course `if (this.parentForm) {}` would also work.

However, this involves a [conversion of `undefined` to a Boolean value](https://www.w3schools.com/js/js_type_conversion.asp).

`this.parentForm !== undefined` checks that the member `this.parentForm` has been initialised, [not performing any type conversion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality). 

So all it does is checking exactly for `undefined` which is a special type in JS. 

`if (myvar) {}` could give unexpected results if `myvar` was actually a Boolean. So if `mayvar` was initialised to `false`, this would not be the same as `undefined`.

The condition should prevent errors in case of non initialised vars.

So in general, I am trying to avoid implicit type conversions.

PR #267, opened 2021-01-26
test(e2e):  improve stability of e2e tests.

tobiasschweizer:
This PR attempts to improve stability of e2e tests.

### DIFF
--- a/e2e/src/app.e2e-spec.ts
+++ b/e2e/src/app.e2e-spec.ts
@@ -35,6 +35,8 @@ describe('Test App', () => {
 
         it('should edit an integer value', async () => {
 
+            const EC = browser.ExpectedConditions;
+
             await page.navigateTo('modify');
 
             let valueEleComp: WebElement = await page.getComponentBySelector('dsp-int-value', timeout);
@@ -42,6 +44,8 @@ describe('Test App', () => {
             const displayEditComp: WebElement = await page.getDisplayEditComponentFromValueComponent(valueEleComp);
 
             await browser.actions().mouseMove(element(by.cssContainingText('.value-component', /^1$/))).perform();
+
+            await browser.wait(EC.presenceOf(element(by.css('.action-bubble'))), timeout, 'Wait for action bubble to be visible.');
 
             expect(element(by.css('.action-bubble')).isDisplayed()).toBeTruthy();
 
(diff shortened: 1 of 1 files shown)


e2e/src/app.e2e-spec.ts
```diff
@@ -43,6 +45,8 @@ describe('Test App', () => {
 
             await browser.actions().mouseMove(element(by.cssContainingText('.value-component', /^1$/))).perform();
 
+            await browser.wait(EC.presenceOf(element(by.css('.action-bubble'))), timeout, 'Wait for action bubble to be visible.');
+
             expect(element(by.css('.action-bubble')).isDisplayed()).toBeTruthy();
```
tobiasschweizer, line 50: Sometimes the tests failed here. Since this only happens from time to time, I have the suspicion it cold be a timing issue when trying to find `.action-bubble`.

mdelez: interesting, let see if it fixes the issue :)

e2e/src/app.e2e-spec.ts
```diff
@@ -71,6 +73,8 @@ describe('Test App', () => {
             const readEle = await page.getReadValueFieldFromValueComponent(valueEleComp);
             expect(await readEle.getText()).toEqual('3');
 
+            // browser.sleep(200000);
```
mdelez, line 76: Should this be left in? I think it's used to keep the browser open when running the tests locally, right?

mdelez (approved):

e2e/src/app.e2e-spec.ts
tobiasschweizer: yes, I find it helpful as a comment so next time I do not have to look for the command

tobiasschweizer:
> interesting, let see if it fixes the issue :)

hard to say since this seems to be intermittent

e2e/src/app.e2e-spec.ts
tobiasschweizer: hard to say since this seems to be intermittent

tobiasschweizer: The strange thing is that `element(by.css('.action-bubble')).isDisplayed()` returned `false` and the assertions was not met. So I am not sure whether the element was not found at all or was found but not displayed.

`presenceOf` makes sure that the element can be found within the timeout, but if the element takes some time to appear, it won't fix that issue.

PR #269, opened 2021-01-27
feat(string-literal-input): updates input on label changes (DSP-1295)

mdelez:
resolves DSP-1295


projects/dsp-ui/src/lib/action/components/string-literal-input/string-literal-input.component.ts
```diff
@@ -210,6 +214,9 @@ export class StringLiteralInputComponent implements OnInit {
         if (!value) {
             value = '';
         }
+        if (!this.form) {
+            return;
+        }
```
kilchenmann, line 219: Is this still necessary? I think we added it because of the wrong implementation of `ngOnChanges`?

mdelez: `updateFormField` will be called from `ngOnChanges` which will be called before the form gets created in `ngOnInit` so I think it's necessary still

kilchenmann (approved):

PR #276, opened 2021-02-05
chore(dependency): update dsp-js to v2.0.0 and dsp-api to v13.2.0

kilchenmann:


mpro7 (approved):

projects/dsp-ui/src/lib/viewer/services/user.service.ts
```diff
@@ -21,8 +22,10 @@ export class UserService {
      *
      * @param userIri the Iri identifying the user.
      */
-    getUser(userIri: string): AsyncSubject<UserResponse> {
-        return this._userCache.getUser(userIri);
+    getUser(userIri: string): Observable<UserResponse> {
+        return this._userCache.getUser(userIri).pipe(
+            map(user => user)
```
tobiasschweizer, line 27: map to the user contained in `UserResponse`?

kilchenmann: This was the solution that worked. But if you (@tobiasschweizer) have time, feel free to checkout this branch and fix it as it should. Thanks

kilchenmann: Btw I wasn't sure about this `AsyncSubject`

tobiasschweizer: ok, I'll take a look at it.

tobiasschweizer: I have substituted all occurrences of `AsyncSubject` for `Observable` in the specs.

Before, the cache returned an `AsyncSubject`, now this is an `Observable`. However, an `AsyncSubject` is also an `Observable`, but now simply `of` can be used in the specs.

About the `map`: we could use it like this:

```ts
    getUser(userIri: string): Observable<ReadUser> {
        return this._userCache.getUser(userIri).pipe(
            map(userRes => userRes.user)
        );
    }
```

So instead of a `UserResponse`, the `ReadUser` would be returned directly.

kilchenmann: Thanks @tobiasschweizer. So, it's solved now? Would you like to "review" and approve?

tobiasschweizer (approved):

PR #190, opened 2020-09-23
DSP-720 - upgrade to Angular 10 + estimation

flavens:
resolves [DSP-720](https://dasch.myjetbrains.com/youtrack/issue/DSP-720)

PR #280, opened 2021-03-18
chore(update angular 10.2) (DSP-720)

tobiasschweizer:
resolves DSP-720

This PR replaces #190 (outdated draft)


tobiasschweizer:
@waychal This went quite well so far. Talk to you tomorrow :-)

waychal:
@tobiasschweizer from what I understood, we normally do not commit `package-lock.json`. It is automatically generated locally. In this case, is it required to push the changes from `package-lock.json` or you accidently added it in commit?

tobiasschweizer:
> @tobiasschweizer from what I understood, we normally do not commit `package-lock.json`. It is automatically generated locally. In this case, is it required to push the changes from `package-lock.json` or you accidently added it in commit?

It was [updated](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json) because of the changes in `package.json` that were then applied by running `npm i`. I think there could always be a [discussion](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5) whether to put `package-lock.json` under version control or not. But since we included it, it has to be updated (Angular version 10 and all deps that changed).

However, this only affects the playground app as the lib's deps are defined in https://github.com/dasch-swiss/dsp-ui-lib/blob/wip/dsp-720-update-angular-10_2/projects/dsp-ui/package.json

waychal:
> > @tobiasschweizer from what I understood, we normally do not commit `package-lock.json`. It is automatically generated locally. In this case, is it required to push the changes from `package-lock.json` or you accidently added it in commit?
> 
> It was [updated](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json) because of the changes in `package.json` that were then applied by running `npm i`. I think there could always be a [discussion](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5) whether to put `package-lock.json` under version control or not. But since we included it, it has to be updated (Angular version 10 and all deps that changed).
> 
> However, this only affects the playground app as the lib's deps are defined in https://github.com/dasch-swiss/dsp-ui-lib/blob/wip/dsp-720-update-angular-10_2/projects/dsp-ui/package.json

Ok. We are not pushing changes for `package-lock.json` in DSP-APP so I though it is same here. Everything else is fine so I am approving the PR.

waychal (approved):

flavens (approved):

PR #288, opened 2021-04-30
fix(viewer): set correct rgba value (DSP-1588)

kilchenmann:
resolves DSP-1588

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/values/color-value/color-value.component.scss
+++ b/projects/dsp-ui/src/lib/viewer/values/color-value/color-value.component.scss
@@ -16,7 +16,7 @@
 }
 
 .color-preview {
-  border: 1px solid rgba(33,33,33.5);
+  border: 1px solid rgba(33,33,33,.5);
   border-radius: 4px;
   padding: 3px 12px;
   margin: 6px 0;


projects/dsp-ui/src/lib/viewer/values/color-value/color-value.component.scss
```diff
@@ -16,7 +16,7 @@
 }
 
 .color-preview {
-  border: 1px solid rgba(33,33,33.5);
+  border: 1px solid rgba(33,33,33,.5);
```
flavens, line 19: .5 or 0.5??

mdelez (approved):
NICE!

projects/dsp-ui/src/lib/viewer/values/color-value/color-value.component.scss
kilchenmann: yes, in scss we don't need a leading 0

kilchenmann: https://github.com/sds/scss-lint/blob/master/lib/scss_lint/linter/README.md#leadingzero

PR #290, opened 2021-05-03
feat(advanced search): try to reuse resource and property selection (DSP-1587)

tobiasschweizer:
resolves DSP-1587

I decided to do this task in two steps:
- move resource class and property selection to a reusable component (this PR)
- use this new component in main form and for linking property specification (upcoming PR)


e2e/src/app.e2e-spec.ts
```diff
@@ -152,27 +154,35 @@ describe('Test App', () => {
...
+            const submitButton4 = await page.getAdvancedSearchSubmitButton(loader);
+
+            expect(await submitButton4.isDisabled()).toBe(true);
 
             const input = await loader.getHarness(MatInputHarness);
 
             await input.setValue('1');
 
-            expect(await submitButton.isDisabled()).toBe(false);
+            const submitButton5 = await page.getAdvancedSearchSubmitButton(loader);
```
tobiasschweizer, line 183: @mdelez I had to add this workaround because the existing instance always returned the old value for the state of the disabled prop.

I believe that we had a similar problem once but there the element was recreated in the DOM which is not the case here. Fishing in the dark here ...

mdelez: yeah from what I can remember, it won't update automatically if the value changes but I think you can reuse the same variable, no?

mdelez (changes requested):
It works quite well on my end! I did notice that when you select either "is equal to" or "is not equal to" for a boolean value, you get two checkboxes. The second checkbox doesn't affect the search in any way. It's just hanging out with its buddy :)

Screen Shot 2021-05-06 at 09 36 44

tobiasschweizer:
> It works quite well on my end! I did notice that when you select either "is equal to" or "is not equal to" for a boolean value, you get two checkboxes. The second checkbox doesn't affect the search in any way. It's just hanging out with its buddy :)
> 
> Screen Shot 2021-05-06 at 09 36 44

Right :) One is for the Boolean value and one is for the sort criterion. There is even an old task for this: https://dasch.myjetbrains.com/youtrack/issue/DSP-354

e2e/src/app.e2e-spec.ts
tobiasschweizer: > but I think you can reuse the same variable, no?

Yes, good point. I think. I could even avoid using a variable in the first place, see https://github.com/dasch-swiss/dsp-ui-lib/pull/290/commits/7e21377cabaf8d5e24b1a5a3beef0a2b512d43e1

This failed locally, let's see how it behaves here.

mdelez:
> > It works quite well on my end! I did notice that when you select either "is equal to" or "is not equal to" for a boolean value, you get two checkboxes. The second checkbox doesn't affect the search in any way. It's just hanging out with its buddy :)
> > Screen Shot 2021-05-06 at 09 36 44
> 
> Right :) One is for the Boolean value and one is for the sort criterion. There is even an old task for this: https://dasch.myjetbrains.com/youtrack/issue/DSP-354

ah, I missed that :)

tobiasschweizer:
> > > It works quite well on my end! I did notice that when you select either "is equal to" or "is not equal to" for a boolean value, you get two checkboxes. The second checkbox doesn't affect the search in any way. It's just hanging out with its buddy :)
> > > Screen Shot 2021-05-06 at 09 36 44
> > 
> > 
> > Right :) One is for the Boolean value and one is for the sort criterion. There is even an old task for this: https://dasch.myjetbrains.com/youtrack/issue/DSP-354
> 
> ah, I missed that :)

maybe I should actually do this task :-)

mdelez (approved):

PR #299, opened 2021-06-08
chore(environment): add test-server config (DSP-1717)

waychal:
resolves DSP-1717

you can run the app with ng s --configuration=test-server or with npm run start-with-test-server


kilchenmann (commented):
Maybe we should add a paragraph in the documentation about this with a comment that it can be used when working on viewer or search module. In some cases it doesn't work (`/read`, `/modify`, etc.) because it has an automatic login for the root user, which is locked on the test server. Maybe here: <https://dasch-swiss.github.io/dsp-ui-lib/how-to-contribute/#run-the-application-locally-in-dev-mode-with-live-reloading>

kilchenmann (approved):

PR #305, opened 2021-06-15
feat(viewer): resolve geoname id and show name (DSP-1212)

tobiasschweizer:
resolves DSP-1212

This PR adds a service that resolves a geoname id.

@gautschr @kilchenmann I'd like to discuss whether and to what degree we could also implement the autocomplete functionality for creating / editing values (like in DSP-Tangoh). As for the backend, we decided not to reimplement the logic with hierarchical lists as it was implemented in SALSAH.

### DIFF
--- a/projects/dsp-ui/src/lib/viewer/services/geoname.service.spec.ts
+++ b/projects/dsp-ui/src/lib/viewer/services/geoname.service.spec.ts
@@ -1,13 +1,20 @@
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
 import { TestBed } from '@angular/core/testing';
-
 import { GeonameService } from './geoname.service';
 
 describe('GeonameService', () => {
   let service: GeonameService;
+  let httpTestingController: HttpTestingController;
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
+    TestBed.configureTestingModule({
+        imports: [
+            HttpClientTestingModule
+        ]
+    });
+
     service = TestBed.inject(GeonameService);
+    httpTestingController = TestBed.inject(HttpTestingController);
   });
 
   it('should be created', () => {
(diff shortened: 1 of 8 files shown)


tobiasschweizer:
As for the token (which is actually more of a username), I made a new entry in the config object. Any user of the app could see it in the network requests to geonames so I thought it does not make much sense to hide it ...

gautschr:
Your suggested autocomplete functionality for creating / editing values would be very comfortable in cases where the data are modified or created. It would eliminate an intermediate time-consuming step where one has to go to geonames to search for the correct ID.

gautschr (approved):
Works perfectly!

tobiasschweizer:
> Your suggested autocomplete functionality for creating / editing values would be very comfortable in cases where the data are modified or created. It would eliminate an intermediate time-consuming step where one has to go to geonames to search for the correct ID.

Ok, I can implement that functionality in a subsequent PR.

mdelez (changes requested):
Should the name of the corresponding Geoname identifier be added to the delete confirmation popup?

Screen Shot 2021-06-16 at 15 23 52

mdelez:
> > Should the name of the corresponding Geoname identifier be added to the delete confirmation popup?
> > Screen Shot 2021-06-16 at 15 23 52
> 
> Good question. Ideally, yes because the name would make it clear to the user what is going to be deleted.
> 
> In practice, there are two problems:
> 
> * the name is not contained in the `ReadValue` itself, but the `ConfirmationDialogComponent` deals with a `ReadValue`
> * the name is obtained asynchronously via the geoname API and displayed using the async pipe. So we don't really know when it's available and delegate this to Angular.
> 
> **EDIT**: the case of a color value is actually similar. The user sees the hex code but not the color ;-)

All good, no need to spend time on it in my mind

mdelez (approved):

PR #307, opened 2021-06-17
feat(viewer): improve editing of geonames (DSP-1212)

tobiasschweizer:
resolves DSP-1212

TODOs:

- [x] make sure geoname id is always represented as a string
- [x] make sure that geonames API returns results in the same language (display and search)
- [x] see if there is a way to preserve the display name when switching to edit mode


gautschr (approved):
works nicely, thank you!

mdelez (approved):
Nice! :D

PR #310, opened 2021-06-23
chore(search): move docs to dsp-app (DSP-1662)

tobiasschweizer:
resolves DSP-1662


kilchenmann:
I don't think that you have to remove those files from the documentation. If someone is using the library only, s/he will need this documentation.

tobiasschweizer:
> I don't think that you have to remove those files from the documentation. If someone is using the library only, s/he will need this documentation.

I thought it'd be good to avoid redundancy. Maybe there could be a link to the user docs?

tobiasschweizer:
I leave this decision to you. You can also close this PR if you'd like to keep the files as they are.

kilchenmann:
I suggest to keep the files

PR #311, opened 2021-06-25
feat(viewer): add checkbox for every resource in list and grid view (DSP-1711)

waychal:
resolves DSP-1711

In this PR, checkbox is added for every resource displayed in list and grid view. It handles 2 cases for both list and grid view:

- When `withMultipleSelection` is `false` (old functionality)
   - if clicked on the single resource, it's details will be displayed

- When `withMultipleSelection` is `true` (new functionality)
  - if clicked on the single resource, it's details will be displayed
  - if checkboxes are selected, to will be redirected to the comparison viewer


projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.ts
```diff
@@ -78,15 +78,15 @@ export class ListViewComponent implements OnChanges {
         this.view = view;
     }
 
-    emitSelectedResource(id: string) {
+    emitSelectedResources(resInfo: object) {
```
kilchenmann, line 81: It would be better to have correct type instead of object.

kilchenmann: Maybe you can create an `Interface` with this object. It will help in the implementation in DSP-APP:

projects/dsp-ui/src/lib/viewer/views/list-view/resource-grid/resource-grid.component.ts
```diff
@@ -20,10 +20,38 @@ export class ResourceGridComponent {
...
 
     constructor() { }
 
+    /**
+     * Maintain the list and count of selected resources
+     *
+     * @param {boolean} checked tells if checkbox is selected
+     * @param {string} resId resource id
+     */
+     viewResourse(checked: boolean, resId: string) {
```
kilchenmann, line 38: typo: `viewResource` instead `viewResourse`

projects/dsp-ui/src/lib/viewer/views/list-view/resource-grid/resource-grid.component.html
```diff
@@ -2,12 +2,12 @@
...
         class="grid-card link"
         [class.selected-resource]="selectedResourceIdx === i"
-        *ngFor="let res of resources.resources; let i = index;"
-        (click)="resourceSelected.emit(res.id)">
+        *ngFor="let res of resources.resources; let i = index;">
         <!-- TODO: add the representation preview here, mat-card-image can be used for images -->
         <mat-card-header class="grid-card-header">
             <mat-card-subtitle class="res-class-label">{{ res.entityInfo.classes[res.type].label }}</mat-card-subtitle>
             <mat-card-title class="res-class-value">{{ res.label }}</mat-card-title>
+            <mat-checkbox id="{{ i }}" (change)="viewResourse($event.checked, res.id)" class="res-checkbox"></mat-checkbox>
```
kilchenmann, line 10: It would be better to have the checkbox in the top right corner (header action area). But at the moment, the grid view is not used in DSP-APP. Maybe we will redesign card later

projects/dsp-ui/src/lib/viewer/views/list-view/resource-grid/resource-grid.component.spec.ts
```diff
@@ -28,7 +29,7 @@ class TestParentComponent implements OnInit {
 
     resources: ReadResourceSequence;
 
-    resIri: string;
+    selectedResources: object;
```
kilchenmann, line 32: same as above: use an `interface` instead of an anonymous object

kilchenmann (changes requested):

waychal:
@kilchenmann I have added the interface and also moved the checkbox to the top right corner. As you said, we can work on the grid template separately to refine it further.

kilchenmann:
> @kilchenmann I have added the interface and also moved the checkbox to the top right corner. As you said, we can work on the grid template separately to refine it further.

I will have a look at it tomorrow morning.

kilchenmann:
Yesterday evening I was thinking about this multiple resource selection. Why not adding an additional `@Input` parameter where the user can say `withMultipleSelection` (true/false). By default it's false. This will help to avoid a breaking change.

waychal:
> Yesterday evening I was thinking about this multiple resource selection. Why not adding an additional `@Input` parameter where the user can say `withMultipleSelection` (true/false). By default it's false. This will help to avoid a breaking change.

@kilchenmann It is good idea. I can update the PR. 
@subotic can you please comment here that it is fine?

subotic:
@waychal yes, it is fine.

waychal:
@kilchenmann I have updated the code to handle both cases of single and multiple resource selection.

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.ts
```diff
@@ -35,16 +58,28 @@ export class ListViewComponent implements OnChanges {
...
+     */
+    @Output() multipleResourcesSelected?: EventEmitter<FilteredResouces> = new EventEmitter<FilteredResouces>();
+
     /**
      * Click on an item will emit the resource iri
      *
      * @param {EventEmitter<string>} resourceSelected
      */
-    @Output() resourceSelected: EventEmitter<string> = new EventEmitter<string>();
+    @Output() singleResourceSelected?: EventEmitter<string> = new EventEmitter<string>();
```
kilchenmann, line 78: Nice implementation, but I suggest to keep the `@Output() resourceSelected: EventEmitter<string> = new EventEmitter<string>();` and mark it in jsDocs as `@deprecated` (<https://jsdoc.app/tags-deprecated.html>). And additional the new line `@Output() singleResourceSelected?: EventEmitter<string> = new EventEmitter<string>();`. In this case we have the new implementation, but the apps will not break.

kilchenmann: Btw. does the `singleResourceSelected` not work when I have `withMultipleSelection = true`. So, I'm wondering how I can open a single resource when clicking on it. Do I have to use `multipleResourcesSelected` in this case? There's no combination of both? I assumed if I click on the resource (and didn't check the checkbox) that I would still be able to view the single resource.

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.scss
```diff
@@ -40,6 +40,6 @@ button.active {
   background-color: $black-12-opacity;
 }
 
-.selected-resource {
+::ng-deep .selected-resource {
```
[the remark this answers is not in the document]
waychal: I agree. We can replace it when there is some alternative available.

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.ts
waychal: > Nice implementation, but I suggest to keep the `@Output() resourceSelected: EventEmitter<string> = new EventEmitter<string>();` and mark it in jsDocs as `@deprecated` (https://jsdoc.app/tags-deprecated.html). And additional the new line `@Output() singleResourceSelected?: EventEmitter<string> = new EventEmitter<string>();`. In this case we have the new implementation, but the apps will not break.

I will update the PR.

> Btw. does the `singleResourceSelected` not work when I have `withMultipleSelection = true`. So, I'm wondering how I can open a single resource when clicking on it. Do I have to use `multipleResourcesSelected` in this case? There's no combination of both? I assumed if I click on the resource (and didn't check the checkbox) that I would still be able to view the single resource.

Yes, you have to use `multipleResourcesSelected` and `withMultipleSelection = true` in case of multiple selection and `singleResourceSelected` and `withMultipleSelection = false` in case of single selection.
 I think it will be confusing for user if he/she clicks on resource to view it and select checkboxes to compare it.

kilchenmann: I think it's also confusing, when a user has to click on the checkbox to open only one resource. The checkboxes should be for multiple selection. As a user I expect to click on a resource to open it directly. Clicking on a checkbox to open it, is not the usual way to open something.

waychal: in `singleResourceSelected`, checkbox is not displayed. User has to click on resource to view it as like we have now.

waychal: In `multipleResourcesSelected`, I agree user needs to select one checkbox which will redirect it to the intermediate page and then display the resource(s). But that's the design we have in Figma now.

If we want user to directly select resource to view it by skipping intermedia page, we need to update the design in Figma.

kilchenmann: In `multipleResourcesSelected` we should have a combination of both. Clicking on a resource will open the single resource. But clicking on the checkboxes it will select 1-n resources and displays the intermediate page. This is a normal use case in UX design.

waychal: I have updated the PR for this.

projects/dsp-ui/src/lib/viewer/views/list-view/resource-list/resource-list.component.html
```diff
@@ -1,26 +1,36 @@
...
-        </p>
-        <h3 matLine class="res-class-value">{{ res.label }}</h3>
+        *ngFor="let resource of resources.resources; let i = index; let last = last">
+        <mat-list-item>
+            <mat-icon matListIcon>image_search</mat-icon>
+            <div matLine (click)="viewResource({ checked: true, resListIndex: i, resId: resource.id, isCheckbox: false})">
+                <p matLine class="res-class-label">
+                    {{ resource.entityInfo.classes[resource.type].label }}
+                </p>
+                <h3 matLine class="res-class-value">{{ resource.label }}</h3>
```
kilchenmann, line 14: Could you set the style for `res-class-label` and `res-class-value` to `margin: 0;`? Otherwise it needs too much space.

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.ts
```diff
@@ -36,15 +60,33 @@ export class ListViewComponent implements OnChanges {
...
+    /**
+     * Click on checkbox will emit the resource info
      *
-     * @param {EventEmitter<string>} resourceSelected
+     * @param {EventEmitter<FilteredResouces>} resourcesSelected
+     */
+    @Output() multipleResourcesSelected?: EventEmitter<FilteredResouces> = new EventEmitter<FilteredResouces>();
+
+    /**
+     * @deprecated
```
kilchenmann, line 75: Please add a description to the `@deprecated` info e.g. Use `singleResourceSelected` instead

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.ts
```diff
@@ -78,15 +120,16 @@ export class ListViewComponent implements OnChanges {
...
-            }
+    // If 'withMultipleSelection' is true, multiple resources are selected for comparision
+    // If 'withMultipleSelection' is false, single resource is selected for viewing
+    emitSelectedResources(resInfo: FilteredResouces) {
+        this.selectedResourceIdx = resInfo.resListIndex;
+
+        if (resInfo.selectionType === "multiple") {
+            this.multipleResourcesSelected.emit(resInfo);
+        } else {
+            this.singleResourceSelected.emit(resInfo.resIds[0]);
```
kilchenmann, line 131: the old/deprecated `resourceSelected` emitter is missing. Please keep `this.resourceSelected.emit(id);` Otherwise it doesn't work anymore in some apps

kilchenmann (commented):
It works. But there are some little details that have to be fixed first

projects/dsp-ui/src/lib/viewer/views/list-view/resource-list/resource-list.component.html
waychal: It is the case only on big screens. `margin:0` removes all the space and it looks crowded. I will make it 'margin: 5px 0', if it ok?

kilchenmann: It's also on small screen to height:
Screen Shot 2021-07-07 at 12 52 45

And this is how it looks with `margin: 0`:
Screen Shot 2021-07-07 at 12 56 34

If you want to add a margin, we should follow to rule of 8. In this case you could set it to `margin: 4px 0;`

waychal: I looks like this on my machine with current style.

image

Anyways I will update the margin.

kilchenmann: Hm... that's weird 😳 it looks different on the two machines. But why?

waychal: Are we using same version? I took this screenshot from `https://admin.test.dasch.swiss/` server.  I did not run the DSP-APP locally to check it there.

For `dsp-ui-lib` server running locally, it looks like below. It does show some spacing. So I assumed that in DSP-APP there are some extra styles applied for this as I did not change this style in this PR until now.

image

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.ts
waychal: done

projects/dsp-ui/src/lib/viewer/views/list-view/list-view.component.ts
waychal: done

projects/dsp-ui/src/lib/viewer/views/list-view/resource-list/resource-list.component.html
kilchenmann: I added the DSP-UI lib from your branch to DSP-APP locally (with yalc) and was running your implementation there. The version on  `https://admin.test.dasch.swiss/` doesn't have your implementation. Of course it looks different 😉

kilchenmann: In your screenshot you have the same big spaces that I have in DSP-APP. This is why I wrote to set the margin to 0 pixels.

waychal:
@kilchenmann @mdelez can you please review the PR.

kilchenmann (approved):
Thanks for all the updates. Works perfect 👍

PR #316, opened 2021-07-15
feat(viewer): split pane to compare multiple resource (DSP-1785)

waychal:
resolves DSP-1785

In this PR, we are splitting pane to show multiple resources selected for comparison.

Note: Every resource displayed on the page has its own slider and in result there are multiple sliders present on the page. It makes the navigation difficult on the page. Does anyone have any idea how we can simplify it?


kilchenmann (approved):
works as expected 👍

PR #322, opened 2021-07-28
chore(deps): bump dsp-js to latest version (DSP-1825)

kilchenmann:
resolves DSP-1825

and bump angular packages to the same version as we use in DSP-APP


kilchenmann:
Btw I thought we can get rid of the jwt in our local storage session object because the DSP-API requests are working with the cookie now. But the sipi uploads still needs the jwt and we need the access to it (described here: <https://docs.dasch.swiss/DSP-API/03-apis/api-v2/editing-values/#upload-files-to-sipi>). So, we have to keep the jwt info in our session object.

mdelez (approved):

PR #329, opened 2021-08-11
feat(display-edit): disable delete depending on cardinality (DSP-1814)

mdelez:
resolves DSP-1814

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "dsp-ui-lib",
       "license": "AGPL-3.0-or-later",
       "dependencies": {
         "@angular/animations": "^11.2.9",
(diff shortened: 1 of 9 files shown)


mdelez:
@kilchenmann should we remove the delete button from the UI in the event the user cannot delete the value or should we disable the button with a hint why it can't be deleted?

kilchenmann:
> @kilchenmann should we remove the delete button from the UI in the event the user cannot delete the value or should we disable the button with a hint why it can't be deleted?

We should disable it. Because the user expects the button. It could be confusing if the button is there in some case but in other cases not.
If the button is disabled the matTooltip will not work. In this case you have to surround (?) it with span tags, where you add the matTooltip

kilchenmann (approved):
LGTM 🚀

PR #331, opened 2021-08-13
feat(action): new progress indicator (just for fun)

kilchenmann:
This PR integrates the new progress indicator depending on our DaSCH logo. You can see an example here: <https://codepen.io/kilchenmann/pen/gOgLqqd>


mdelez (approved):
The codepen example doesn't seem to work for me, it's just a constant loading animation (I'm kidding). Looks great!

